PR #2, opened 2020-07-27
Added pyaudio as an input source

will-rice:
This change introduces the ability to accept microphone input via PyAudio and establishes the base SpeechConfig class.


.circleci/config.yml
```diff
@@ -8,6 +8,7 @@ jobs:
 
         steps:
             - checkout
+            - run: sudo apt-get install portaudio19-dev python3-pyaudio
```
brentspell, line 11: you're using a virtualenv and installing pyaudio with pip, do you need to install python3-pyaudio?

tests/mic/test_audio_input.py
```diff
@@ -0,0 +1,25 @@
...
+    config = SpeechConfig(params)
+    mic = PyAudioMicrophoneInput(config)
+    mic.audio = mock.MagicMock()
+    mic.build()
+    # start audio stream
+    mic.start()
+    assert mic.is_active()
+    # read a single frame
+    frame = mic.read()
+    assert frame
```
brentspell, line 22: it would be good to verify that you get the expected number of samples from this call

spokestack/mic/pyaudio.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Retrieve microphone input with PyAudio
+
+    Args:
+        config (SpeechConfig): configuration paramaters
+    """
+
+    def __init__(self, config: SpeechConfig) -> None:
+
+        self.config = config
```
brentspell, line 20: for a library like this, it would be better to make these attributes private (_ prefix)
for properties that you want to expose to the client, make the attribute private and create an @property method that provides read-only access

spokestack/mic/pyaudio.py
```diff
@@ -0,0 +1,51 @@
...
+    def read(self) -> bytes:
+        return self._stream.read(self.frame_size, exception_on_overflow=False)
+
+    def start(self) -> None:
+        self._stream.start_stream()
+
+    def stop(self) -> None:
+        self._stream.stop_stream()
+
+    def is_active(self) -> bool:
```
brentspell, line 44: is_active and is_stopped would make more sense as @property methods

spokestack/config.py
```diff
@@ -0,0 +1,12 @@
+"""
+This module contains the SpeechConfig object which
+sets the parameters for the SpeechPipeline
+"""
+
+
+class SpeechConfig:
+    """ Speech Configuration """
+
+    def __init__(self, config: dict) -> None:
```
brentspell, line 10: configuration dictionaries make sense in the java world, but they aren't very pythonic
it would make more sense to use kwargs on the SpeechConfig class to set up the configuration properties
another option would be to get rid of SpeechConfig altogether and use kwargs on the constructors of the components that actually need configuration values

spokestack/mic/pyaudio.py
```diff
@@ -0,0 +1,85 @@
...
+            input=True,
+            input_device_index=device["index"],
+            format=pyaudio.paInt16,
+            frames_per_buffer=self._frame_size,
+            channels=1,
+            rate=self._sample_rate,
+            start=False,
+        )
+
+    def __call__(self) -> bytes:
```
brentspell, line 37: I think making speech pipeline components callable makes sense, but not for inputs like this
`mic.read()` reads much better than `mic()` and fits most the expectation for the method to call on an input object

tests/mic/test_audio_input.py
```diff
@@ -0,0 +1,23 @@
...
+from unittest.mock import patch
+
+from spokestack.mic.pyaudio import PyAudioMicrophoneInput
+
+
+@patch("spokestack.mic.pyaudio.pyaudio")
+def test_audio_input(mock_class):
+    """ Test for microphone input """
+    config = {"sample_rate": 16000, "frame_width": 10}
+    mic = PyAudioMicrophoneInput(**config)
```
brentspell, line 13: instead of passing a dictionary, it would be simpler just to pass the kwargs directly here

tests/mic/test_audio_input.py
```diff
@@ -0,0 +1,23 @@
...
+Mock tests for microphone input
+"""
+from unittest.mock import patch
+
+from spokestack.mic.pyaudio import PyAudioMicrophoneInput
+
+
+@patch("spokestack.mic.pyaudio.pyaudio")
+def test_audio_input(mock_class):
+    """ Test for microphone input """
```
brentspell, line 11: tests generally don't require docstrings, because they wouldn't be included in the api reference
is black or something requiring docstrings for tests?

spokestack/mic/pyaudio.py
```diff
@@ -0,0 +1,85 @@
...
+    def stopped(self) -> bool:
+        """ Stream stopped property
+
+        :return: 'True' if stream is stopped, 'False' otherwise
+        :rtype: bool
+        """
+        return self._stream.is_stopped()
+
+    @property
+    def sample_rate(self):
```
brentspell, line 78: if you're going to make all these properties public (and not every attribute needs to be exposed as a public property), then they should have tests - check your coverage numbers for this module

spokestack/mic/pyaudio.py
will-rice: Yeah I agree. I don't know why that seemed like a good idea yesterday for this class.

tests/mic/test_audio_input.py
will-rice: It's not black. I think it is pylint with a bad configuration so it is inspecting tests.

PR #3, opened 2020-08-01
Added Voice Activity Detection

will-rice:
This changes introduces voice activity detection from webrtcvad and the SpeechContext class.

### DIFF
--- a/spokestack/vad/webrtc.py
+++ b/spokestack/vad/webrtc.py
@@ -8,22 +8,25 @@ import webrtcvad  # type: ignore
 from spokestack.context import SpeechContext
 
 
-MODE = {"quality": 0, "low_bitrate": 1, "aggresive": 2, "very_aggresive": 3}
+QUALITY = 0
+LOW_BITRATE = 1
+AGGRESSIVE = 2
+VERY_AGGRESSIVE = 3
 
 
 class VoiceActivityDetector:
     """ This class detects the presence of voice in a frame of audio.
 
-    :param sample_rate: sample rate of the audio
+    :param sample_rate: sample rate of the audio (Hz)
     :type sample_rate: int
-    :param frame_width: width of the audio frame
+    :param frame_width: width of the audio frame: 10, 20, or 30 (ms)
     :type frame_width: int
-    :param vad_rise_delay: rise delay
+    :param vad_rise_delay: rising edge delay (ms)
     :type vad_rise_delay: int
-    :param vad_fall_delay: fall delay
+    :param vad_fall_delay: falling edge delay (ms)
     :type vad_fall_delay: int
-    :param mode: mode for vad
-    :type mode: str
+    :param mode: named constant to set mode for vad
+    :type mode: int
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


tests/vad/test_webrtc.py
```diff
@@ -0,0 +1,36 @@
...
+    context = SpeechContext()
+    detector = VoiceActivityDetector(
+        sample_rate=16000,
+        frame_width=10,
+        vad_rise_delay=0,
+        vad_fall_delay=20,
+        mode="quality",
+    )
+
+    # test detection
```
brentspell, line 23: this only tests that the vad is eventually triggered because the mock is always returning true, you'll also want to test the following states: 
. verify that the vad remains untriggered if vad_rise_delay hasn't been reached and the mock is returning true
. verify that the vad remains triggered if vad_fall_delay hasn't been reached and the mock is returning false
. verify that the vad becomes untriggered if vad_fall_delay has been reached and the mock is returning false

tests/vad/test_webrtc.py
```diff
@@ -0,0 +1,36 @@
...
+    frame = np.zeros(160, np.int16).tobytes()
+    detector(context, frame)
```
brentspell, line 25: you're testing two calls here to trigger the vad, but you have vad_rise_delay set to 0 above, so it should actually becoming triggered after the first call

tests/vad/test_webrtc.py
```diff
@@ -0,0 +1,36 @@
...
+    assert detector.run_value != 0
```
brentspell, line 26: `run_value` and `run_length` are implementation details, so they should be private, and the test shouldn't inspect them
clients and the test only care about the state of the is_speech indicator

spokestack/vad/webrtc.py
```diff
@@ -0,0 +1,74 @@
...
+This module contains the webrtc component for
+voice activity detection (vad)
+"""
+
+import webrtcvad  # type: ignore
+
+from spokestack.context import SpeechContext
+
+
+MODE = {"quality": 0, "low_bitrate": 1, "aggresive": 2, "very_aggresive": 3}
```
brentspell, line 11: this would be simpler if you created named constants QUALITY, LOW_BITRATE, AGGRESSIVE, VERY_AGGRESSIVE
then the client could pass symbolic constants instead of strings and you wouldn't need to map them to webrtc's values

spokestack/vad/webrtc.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class VoiceActivityDetector:
+    """ This class detects the presence of voice in a frame of audio.
+
+    :param sample_rate: sample rate of the audio
```
brentspell, line 17: it's good to specify the units for values like this, like "sample rate of the audio, in Hz"
ditto for frame_width, vad_rise_delay, vad_fall_delay (milliseconds)

spokestack/vad/webrtc.py
```diff
@@ -0,0 +1,74 @@
...
+    :type sample_rate: int
+    :param frame_width: width of the audio frame
+    :type frame_width: int
+    :param vad_rise_delay: rise delay
+    :type vad_rise_delay: int
+    :param vad_fall_delay: fall delay
```
brentspell, line 23: this is not a descriptive comment - it's just rewriting the parameter name with a space

spokestack/vad/webrtc.py
```diff
@@ -0,0 +1,74 @@
...
+import webrtcvad  # type: ignore
+
+from spokestack.context import SpeechContext
+
+
+MODE = {"quality": 0, "low_bitrate": 1, "aggresive": 2, "very_aggresive": 3}
+
+
+class VoiceActivityDetector:
+    """ This class detects the presence of voice in a frame of audio.
```
brentspell, line 15: it would be good to generate the documentation for this module, then read it and see if it should be fleshed out more - think about whether this would be useful documentation for a newcomer to the library

spokestack/vad/webrtc.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(
+        self,
+        sample_rate: int,
+        frame_width: int,
+        vad_rise_delay: int,
+        vad_fall_delay: int,
+        mode: str,
+    ) -> None:
+
+        self.sample_rate = sample_rate
```
brentspell, line 38: all of these member variables should probably be private

spokestack/vad/webrtc.py
will-rice: that makes so much sense. thanks for pointing that out.

spokestack/vad/webrtc.py
will-rice: can I do this with the sphinx PR?

brentspell: I don't see why not

PR #4, opened 2020-08-04
Added Documentation via Sphinx

will-rice:
This change introduces a way to build auto-docs from function/class docstrings with Sphinx. The following were added:
- A directory containing the documentation called `docs`
- Google style docstrings that can be converted to reST with napoleon


will-rice:
ok that makes a lot of sense. I’ll do that.

brentspell:
I forget where we're at on this one, is it ready to merge?

will-rice:
it is ready.

PR #5, opened 2020-08-20
Addition of Wakeword Module

will-rice:
This change brings in the ability to use spokestack's wakeword technology in python. Some of the notable additions are:

- TFLiteModel: The base class for serving tflite models with support for  multiple inputs and outputs.
- RingBuffer: A multi-dimensional circular queue implemented with numpy.
- WakewordDetector: The module that detects the presence of a wakeword in an audio stream.


spokestack/models/tensorflow.py
```diff
@@ -25,18 +25,21 @@ def __init__(self, model_path: str, **kwargs: Any) -> None:
         self._output_details = self._interpreter.get_output_details()
         self._interpreter.allocate_tensors()
 
-    def __call__(self, inputs: List[np.ndarray]) -> List[np.ndarray]:
```
brentspell: a simpler way to do this would be to accept the parameter as *args, then it's always a list, and clients can either pass the input tensors directly or as *list: 
```
model(tensor1, tensor2)
model(*tensors)
```

spokestack/wakeword/ring_buffer.py
```diff
@@ -110,16 +109,21 @@ def read(self) -> np.ndarray:
...
+    def read_all(self) -> np.ndarray:
         """ Dumps the entire contents of the buffer to an array
 
         Returns: Array with full contents of the buffer
 
         """
-        return self._buffer[: self.capacity]
+
+        self.rewind()
+        full = np.empty_like(self._buffer[: self.capacity])
```
brentspell, line 124: this assumes that the buffer is full when you read it
if the buffer isn't full, you'll return a numpy array with the whole capacity, and the client would have to filter it for empties to determine the number of items read
a simpler way to do this would be to read the buffer elements into a list, then call `np.stack` to concatenate them
also, the buffer class has a dtype parameter, so it would be better to convert the results to that instead of `np.float32`

spokestack/wakeword/tflite.py
```diff
@@ -59,17 +66,18 @@ def __init__(
...
-            shape=[1, int(self.window_size / 2 + 1)]
-        )
+        self.fft_frame = np.empty((1, int(self.window_size / 2 + 1)), np.float32)
         self.frame_window: RingBuffer = RingBuffer(
             shape=[self.mel_length, self.mel_width]
         )
         self.encode_window: RingBuffer = RingBuffer(
             shape=[self.encode_length, self.encode_width]
         )
+        self.frame_window.fill(np.zeros((1, self.mel_width), np.float32))
```
brentspell, line 79: it would be simpler if the buffer could just be filled with a scalar value, instead of having to create a vector

spokestack/wakeword/tflite.py
```diff
@@ -94,87 +103,46 @@ def __call__(self, context: SpeechContext, frame) -> None:
...
-
-        Args:
-            context (SpeechContext): current state of the speech pipeline
-            frame (np.ndarray): a single frame of an audio signal
+    def _sample(self, context: SpeechContext, frame) -> None:
 
-        Returns: None
+        frame /= 2 ** 15 - 1
+        frame = np.clip(frame, -1.0, 1.0)
+        frame = np.append(frame[0], frame[1:] - self.pre_emphasis * frame[:-1])
```
brentspell, line 118: `self._prev_sample` should be prepended to the frame here, in order to maintain sample continuity from frame to frame, something like this:

```
prev_sample = frame[-1]
frame -= self.pre_emphasis * np.append(self._prev_sample, frame[:-1])
self._prev_sample = prev_sample
```

spokestack/wakeword/tflite.py
```diff
@@ -59,17 +66,18 @@ def __init__(
         )
         self.encode_width: int = wake_encode_width
         self.state_width: int = wake_encode_width
+        self._hann_window = np.hanning(self.window_size)
 
         self.sample_window: RingBuffer = RingBuffer(shape=[self.window_size])
-        self.fft_window: RingBuffer = RingBuffer(
-            shape=[1, int(self.window_size / 2 + 1)]
-        )
+        self.fft_frame = np.empty((1, int(self.window_size / 2 + 1)), np.float32)
```
brentspell, line 72: you shouldn't need a member for this, just pass the frame from the `_analyze` method to the `_filter` method when you call it

spokestack/wakeword/tflite.py
```diff
@@ -94,87 +103,46 @@ def __call__(self, context: SpeechContext, frame) -> None:
...
-        frame = self.sample_window.to_array()
-        window = signal.get_window("hann", self.window_size)
-        frame = np.fft.rfft(frame * window, n=self.window_size)
+        frame = self.sample_window.read_all()
+        frame = np.fft.rfft(frame * self._hann_window, n=self.window_size)
         frame = np.abs(frame)
-        self.sample_window.reset()
 
-        self.fft_window.write(frame)
+        self.fft_frame[0] = frame
```
brentspell, line 134: no need for a member/assignment here, the `frame` variable can just be passed to the `_filter` function

spokestack/wakeword/tflite.py
```diff
@@ -204,14 +163,8 @@ def _detect(self, context: SpeechContext) -> None:
     def reset(self) -> None:
         """ Resets the currect WakewordDetector state """
         self.sample_window.reset()
-        self.frame_window.reset().fill(0.0)
-        self.encode_window.reset().fill(0.0)
-        self.state = np.zeros(self.encode_model.input_details[1]["shape"], np.float32)
-        self.fft_window.reset()
+        self.frame_window.reset().fill(np.zeros((1, self.mel_width), np.float32))
```
brentspell, line 166: same note here as above, consider just passing the scalar 0 to fill the buffer, not shapes/dtypes needed

spokestack/wakeword/tflite.py
will-rice: yeah I think the numpy `.fill` will be a better way of accomplishing this.

spokestack/wakeword/tflite.py
```diff
@@ -101,67 +111,70 @@ def __call__(self, context: SpeechContext, frame) -> None:
...
                 if context.is_speech:
                     self._analyze(context)
-                # prepare the sample window for the next iteration
                 self.sample_window.rewind().seek(self.hop_length)
 
     def _analyze(self, context: SpeechContext) -> None:
-        # read the full contents of the sample window
+        # read the full contents of the sample window to calculate a single frame
+        # of the STFT by applying the DFT to a real-valued input and
+        # polarize the output by computing the absolute value
```
brentspell, line 138: I wouldn't use the term polarize here - something like "take the magnitude of the complex DFT"

PR #6, opened 2020-08-27
First Draft

will-rice:


spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,106 @@
...
+                "rate": sample_rate,
+                "language": language,
+                "limit": limit,
+            }
+        )
+
+        self._socket_url: str = socket_url
+        self._key_id: str = key_id
+        self._key: bytes = key_secret.encode("utf-8")
+        signature = hmac.new(
```
brentspell, line 38: this is fine for now and I know the android version does this, but I don't really like pre-calculating the request signature like this
at some point, we will add things like timestamps for replay attacks, and that will require rewriting all of this to calculate the request signature correctly before it is submitted

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,106 @@
...
+        if not self._connected:
+            raise ConnectionError("Not Connected")
+
+        message = {
+            "keyId": self._key_id,
+            "signature": self._signature,
+            "body": self._body,
+        }
+        self._socket.send(json.dumps(message))
+        self._socket.recv()
```
brentspell, line 75: `self._is_final` should be reset here to false, since the object can be reused

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,106 @@
...
+        return self._connected
+
+    def close(self) -> None:
+        self._socket.close()
+
+    def connect(self) -> None:
+        if self._connected:
+            raise ConnectionError("Already Connected")
+        self._socket.connect(f"{self._socket_url}/v1/asr/websocket")
+        self._connected = True
```
brentspell, line 63: you can probably eliminate the `_connected` member variable and just check `self._socket` for `None` to detect the connected state
this will require moving the `WebSocket()` call in here from the constructor and initializing `_socket` to None there

spokestack/asr/cloud_client.py
brentspell, line 75: it would probably be better to convert api error responses to exceptions here (and in the receive function), in order to avoid silent failure for applications
I would create a new exception class and use it to encapsulate failed responses

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,106 @@
...
+            pass
+        self._socket.timeout = timeout
+
+    @property
+    def response(self) -> dict:
+        return self._response
+
+    @property
+    def is_final(self):
+        return self._is_final
```
brentspell, line 106: I think it still makes sense to provide a non-streaming method in this class, where you can just pass an audio buffer/array and get the final transcript

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(
+        self,
+        spokestack_id: str = "",
+        spokestack_secret: str = "",
+        language: str = "en",
+        sample_rate: int = 16000,
+        frame_width: int = 10,
+    ) -> None:
+        self._client_id: str = spokestack_id
```
brentspell, line 22: no need to store `spokestack_id`, `secret`, `sample_rate`, `frame_width` here

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+        self._secret: str = spokestack_secret
+        self._language: str = language
+        self._sample_rate: int = sample_rate
+        self._frame_width: int = frame_width
+        self.max_idle_count = 5000 / frame_width
```
brentspell, line 27: `idle_timeout` should probably be configurable

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+        self._client: CloudClient = CloudClient(
+            self._client_id, self._secret, self._language, sample_rate=self._sample_rate
+        )
+        self._is_active = False
+        self._idle_count: int = 0
+        self._final = True
```
brentspell, line 33: no need to store `_final`, since you can get it from the client

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+        self._context: SpeechContext
```
brentspell, line 34: we don't want to store the context, use the value that is passed to the callable

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+        self._idle_count = 0
+
+    def _send(self, frame) -> None:
+        self._client.send(frame)
+
+    def _receive(self):
+        self._client.receive()
+        self._on_message(self._client.response)
+
+    def _on_message(self, message):
```
brentspell, line 70: this method can be merged with `_receive`

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+        hypotheses = message.get("hypotheses")
+        for hypothesis in hypotheses:
+            self._context.transcript = hypothesis.get("transcript")
```
brentspell, line 73: this will overwrite the context's transcripts with the lowest scoring transcript in the response
you just want one transcript, so take the first one in the list and assign it to the context, but also keep in mind that the list may be empty

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,106 @@
...
+    @property
+    def is_connected(self) -> bool:
+        return self._connected
+
+    def close(self) -> None:
+        self._socket.close()
+
+    def connect(self) -> None:
+        if self._connected:
+            raise ConnectionError("Already Connected")
```
brentspell, line 61: you could probably be a bit friendly here to the caller and silently ignore duplicate connect calls

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+            self._commit()
+        elif not self._final:
+            self._receive()
+        elif self._idle_count < self.max_idle_count:
+            self._idle_count += 1
+        else:
+            self._client.disconnect()
+
+    def _begin(self) -> None:
+        if not self._client.is_connected:
```
brentspell, line 57: if the client ignores duplicate connect calls, you can remove this check

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,89 @@
...
+            self._context.transcript = hypothesis.get("transcript")
+            self._context.confidence = hypothesis.get("confidence")
+        self._final = self._client.response.get("final")
+
+    def _commit(self) -> None:
+        self._is_active = False
+        self._client.end()
+
+    def reset(self) -> None:
+        if self._client.is_connected:
```
brentspell, line 82: the client ignores duplicate connect calls, so you can remove this check

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,106 @@
...
+        self._socket.send(json.dumps(message))
+        self._socket.recv()
+
+    def disconnect(self) -> None:
+        if self._connected:
+            self._socket.close()
+        self._connected = False
+
+    def send(self, frame):
+        self._socket.send_binary(frame)
```
brentspell, line 83: you probably want a check to see if you are connected here, in order to avoid a null call - same for `end` and `receive`

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(
+        self,
+        socket_url: str,
+        key_id: str,
+        key_secret: str,
+        audio_format: str = "PCM16LE",
+        sample_rate: int = 16000,
+        language: str = "en",
+        limit: int = 10,
+        idle_timeout: Any = None,
```
brentspell, line 38: this should be union of `None` or `float`
for example, strings should not be accepted here

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+            "final": False,
+            "hypotheses": [],
+            "status": None,
+        }
+        self._sample_rate: int = sample_rate
+        self._is_final: bool = False
+        self._idle_timeout = idle_timeout
+        self._idle_count: int = 0
+
+    def __call__(self, audio: np.ndarray, n_best: int = 1) -> List[str]:
```
brentspell, line 69: `n_best` here should be consistently named with `limit` above, I recommend `limit`

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+
+        Args:
+            audio (np.ndarray): np.int16 array of audio
+            n_best (int): number of predictions to return
+
+        Returns: n_best list of transcripts
+
+        """
+
+        chunk_size = 2 * int(0.01 * self._sample_rate)
```
brentspell, line 80: this chunk size is too small for non-streaming audio, so you'll have a lot more overhead
a good size for bulk recognition would probably be 1 second of audio (32kb)
also, you are slicing a 16-bit array (not bytes), so multiplying by 2 here doesn't make sense

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+        self.initialize()
+
+        for i in range(0, len(audio), chunk_size):
+            frame = audio[i:][:chunk_size]
+            self.send(frame)
+            self.receive()
+
+        self.end()
+        while not self._is_final:
+            if self._idle_timeout and (self._idle_count > self._idle_timeout):
```
brentspell, line 91: idle checks don't make sense for bulk recognition, just wait until `is_final` or an error occurs

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+                break
+            else:
+                self.receive()
+                self._idle_count += 1
+        self.disconnect()
+
+        transcripts = []
```
brentspell, line 98: the confidence values should be returned along with the transcripts

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+        self._sample_rate: int = sample_rate
+        self._is_final: bool = False
+        self._idle_timeout = idle_timeout
+        self._idle_count: int = 0
+
+    def __call__(self, audio: np.ndarray, n_best: int = 1) -> List[str]:
+        """ Audio to text interface for the cloud client
+
+        Args:
+            audio (np.ndarray): np.int16 array of audio
```
brentspell, line 73: you're requiring that the array be `np.int16`, so it would be good to verify the dtype here
alternatively, you could convert from float32, bytes, etc. automatically based on the type

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+    @property
+    def is_connected(self) -> bool:
+        """ status of the socket connection """
+        if self._socket:
+            return True
+        return False
+
+    def close(self) -> None:
+        """ closes socket """
+        self._socket.close()
```
brentspell, line 115: since you already have a `disconnect` method, I would just remove the `close` method
if you choose to keep it, you should also set the socket to `None` so that `is_connected` returns `False`

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def connect(self) -> None:
+        """ connects to websocket """
+        if self._socket:
+            pass
```
brentspell, line 120: it would be simpler to write `if self._socket is None` instead of using `pass` here

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+            raise APIError(self._response)
+
+    def disconnect(self) -> None:
+        """ disconnects client socket connection """
+        if self._socket:
+            self.close()
+        self._socket = None
+
+    def send(self, frame):
+        """ sends a single frame of binary audio """
```
brentspell, line 148: `frame` should have a mypy type, and the docstring should indicate this
also, it looks like the value that is being passed in here is a `np.array`, so the docstring shouldn't indicate that the value is a binary
and I'm guessing that the websocket stuff is transparently converting the array to bytes, but it would be better to make this explicit by calling `to_bytes` here

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,92 @@
...
+        self._client: CloudClient = CloudClient(
+            spokestack_id,
+            spokestack_secret,
+            language,
+            sample_rate=sample_rate,
+            idle_timeout=int(idle_timeout / frame_width),
+        )
+        self._is_active = False
+
+    def __call__(self, context: SpeechContext, frame: np.ndarray) -> None:
```
brentspell, line 43: no docstring

spokestack/asr/cloud_client.py
```diff
@@ -0,0 +1,206 @@
...
+        return self._idle_count
+
+    @idle_count.setter
+    def idle_count(self, value: int):
+        """ sets the idle counter"""
+        self._idle_count = value
+
+
+class APIError(Exception):
+    """ Spokestack api error pass through """
```
brentspell, line 203: `response` isn't documented as a parameter in the docstring

spokestack/asr/speech_recognizer.py
```diff
@@ -0,0 +1,92 @@
...
+            context.confidence = hypothesis["confidence"]
+        self._is_final = self._client.response.get("final")
+
+    def _commit(self) -> None:
+        self._is_active = False
+        self._client.end()
+
+    def reset(self) -> None:
+        """ resets client connection """
+        if self._client.is_connected:
```
brentspell, line 84: there is no need to check `is_connected` here, because the client does so

spokestack/asr/cloud_client.py
```diff
@@ -66,18 +66,30 @@ def __init__(
...
+            limit (int): number of predictions to return
 
-        Returns: n_best list of transcripts
+        Returns: list of transcripts, and their confidence values of size limit
 
         """
+        if isinstance(audio, bytes):
+            audio = np.frombuffer(audio, np.int16)
+        elif np.issubdtype(audio.dtype, np.int16):
+            audio = audio
```
brentspell, line 85: no need to assign audio to itself here, just put this as the last clause and invert it

spokestack/asr/cloud_client.py
```diff
@@ -140,14 +137,17 @@ def initialize(self) -> None:
 
     def disconnect(self) -> None:
         """ disconnects client socket connection """
-        if self._socket:
-            self.close()
         self._socket = None
```
brentspell, line 140: for an explicit disconnect, you still want to call `socket.close` instead of waiting for the garbage collector to do it

spokestack/asr/cloud_client.py
```diff
@@ -140,14 +137,17 @@ def initialize(self) -> None:
...
 
-    def send(self, frame):
-        """ sends a single frame of binary audio """
+    def send(self, frame: np.ndarray):
+        """ sends a single frame of audio
+
+        Args:
+            frame (np.ndarray):
```
brentspell, line 146: no description

spokestack/asr/cloud_client.py
```diff
@@ -137,13 +135,14 @@ def initialize(self) -> None:
 
     def disconnect(self) -> None:
         """ disconnects client socket connection """
+        self._socket.close()
```
brentspell, line 138: what happens if the client is already disconnected?
this would be a good test to write

will-rice: attribute error because it is `None`. I shouldn't let it run `disconnect` again.

will-rice: well. I mean check if the socket `is_connected` before calling close again

brentspell (approved):
lgtm

PR #8, opened 2020-09-08
Addition of Spokestack NLU and Slot Parsing

will-rice:
The following introduces Spokestack's NLU with the ability to parse slots correctly classified by the model. The major additions are:

- TFLiteNLU class which serves the nlu model
- Parsers for the supported Spokestack types: entity, selset, integer, digit


brentspell:
might be best to get @space-pope to look at this one

spokestack/nlu/parsers/integer.py
```diff
@@ -0,0 +1,70 @@
+"""
+This module contains the logic to parse integers from NLU results
+"""
+from typing import Any, Dict, Union
+
+from spokestack.nlu.parsers import maps
+from spokestack.nlu.parsers.digits import DIGIT_SPLIT_RE
```
space-pope, line 7: I wouldn't have the parsers import from each other; seems like the word maps and this regex belong in either a utils module or `parsers.__init__`.

spokestack/nlu/parsers/integer.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def parse(metadata: Dict[str, Any], raw_value: str) -> Union[int, None]:
+    """ Parser for Integer slots
```
space-pope, line 11:
```suggestion
    """ Integer Parser
```
^ just changed this to match the others, though I think I might write them all as, e.g., "integer slot parser"

spokestack/nlu/parsers/integer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    parsed_values = []
+    for token in tokens:
+        try:
+            parsed = int(token)
+            parsed_values.append(parsed)
+        except ValueError:
+            reduced = _parse_reduce(token, parsed_values)
+            if not reduced:
+                return None
```
space-pope, line 32:
```suggestion
            if _parse_reduce(token, parsed_values) is None:
                return None
```
not mandatory, just a thought

spokestack/nlu/parsers/integer.py
```diff
@@ -0,0 +1,70 @@
...
+def _collapse(multiplier, so_far):
+    collapsed = []
+    total = 0
+    for number in so_far:
+        if number > multiplier:
+            collapsed.append(number)
+        else:
+            total += number
+    if not total > 0:
+        total = 1
```
space-pope, line 64:
```suggestion
    total = max(total, 1)
```

spokestack/nlu/parsers/maps.py
```diff
@@ -0,0 +1,130 @@
...
+
+ENG_EXP10: Dict[str, int] = {"hundred": 2, "thousand": 3}
+
+MULTIPLIERS: Dict[str, int] = {
+    "hundred": 100,
+    "thousand": 1000,
+    "million": int(1e6),
+    "billion": int(1e9),
+}
+WORD_TO_NUM: Dict[str, int] = {
```
space-pope, line 66: You've got `MULTIPLIERS` in here already, but you could make it a bit shorter by `**`ing `ENG_ZERO`, `ENGMOD10`, and `ENG_MOD20` and explicitly specifying what's left.

spokestack/nlu/parsers/maps.py
```diff
@@ -0,0 +1,130 @@
...
+    "fifty": 50,
+    "fifdee": 50,
+    "fiddy": 50,
+    "fiftie": 50,
+    "sixty": 60,
+    "sixtie": 60,
+    "sixdee": 60,
+    "seventy": 70,
+    "seventie": 70,
+    "sevendie": 70,
```
space-pope, line 122: Have you seen the "die"/"dee" variants as ASR results, or are you aiming for exceptionally bad chat typing? In general, we've opted to include homonyms, but not typos.

spokestack/nlu/parsers/selset.py
```diff
@@ -0,0 +1,27 @@
+"""
+This module contains the logic to parse selsets from NLU results
```
space-pope, line 2: more doc—remember how unintuitive the selset concept can be.

spokestack/nlu/tflite.py
```diff
@@ -0,0 +1,119 @@
+"""
+This module contains the class to serve TFLite NLU models
```
space-pope, line 2: doc

spokestack/nlu/tflite.py
```diff
@@ -0,0 +1,119 @@
...
+            i: intent["name"] for i, intent in enumerate(self._metadata["intents"])
+        }
+        self._tag_decoder = {i: tag for i, tag in enumerate(self._metadata["tags"])}
+        self._intent_meta = {
+            intent.pop("name"): intent for intent in self._metadata["intents"]
+        }
+        self._warm_up()
+
+    def __call__(self, utterance: str) -> Dict[str, Any]:
+        """ Forward Pass
```
space-pope, line 39: Technically I suppose inference is a forward pass, but isn't this terminology mainly for training? I'd elaborate here a bit. This classifies an utterance, returning the intent predicted by the model, its confidence in that prediction, and a dict of slots recognized in the utterance and parsed according to types specified in the model metadata.

spokestack/nlu/tflite.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def _decode(self, outputs) -> Tuple[str, List[str], float]:
+        # to get the index of the highest probability we
+        # apply argmax to the posteriors which allows the
+        # labels to be decoded with an integer to string mapping
+        # we derive the confidence from the highest probability
+        intent_posterior, tag_posterior = outputs
+        intents = self._decode_intent(intent_posterior)
+        tags = self._decode_tags(tag_posterior)
+        confidence = np.max(intent_posterior)
```
space-pope, line 104: This extra trip through the posteriors wouldn't be necessary if you let `_decode_intent` return both the index and value of the argmax.

spokestack/nlu/tflite.py
```diff
@@ -0,0 +1,119 @@
...
+        return [self._tag_decoder.get(tag) for tag in tags]
+
+    def _decode_intent(self, posterior):
+        intent = np.argmax(posterior, -1)[0]
+        return self._intent_decoder.get(intent)
+
+    def _parse_slots(self, slot_meta, slots):
+        slot_type = slot_meta["type"]
+        parser = import_module(f"spokestack.nlu.parsers.{slot_type}")
+        facets = json.loads(slot_meta["facets"])
```
space-pope, line 118: It would be best to do this JSON loading only once if possible, caching the results somewhere.

tests/nlu/parsers/test_digits.py
```diff
@@ -0,0 +1,35 @@
+"""
+This module contains the tests for the digits parser
+"""
+from spokestack.nlu.parsers import digits
+
+
+def test_digits():
```
space-pope, line 7: Check the Android version's tests here; it'd be good to have some sad paths here (unparsable input beyond the empty string, situations where the `count` metadata is violated, etc). Also things that exercise normalization and maybe some mixed numeric/word input.

tests/nlu/parsers/test_entity.py
```diff
@@ -0,0 +1,11 @@
+"""
+This module contains the tests for the entity parser
+"""
+from spokestack.nlu.parsers import entity
+
+
+def test_entity():
```
space-pope, line 7: test for empty string?

tests/nlu/parsers/test_integer.py
```diff
@@ -0,0 +1,33 @@
...
+
+    metadata = {"range": [1, 500]}
+    raw_value = "four hundred four"
+    parsed = integer.parse(metadata, raw_value)
+    assert parsed == 404
+
+    metadata = {"range": [1, 500]}
+    raw_value = "four hundred hundred"
+    parsed = integer.parse(metadata, raw_value)
+    assert not parsed
```
space-pope, line 33: It's sort of unclear just from looking at the test whether this fails because it's out of range or because it's an ill-formed number. You'll also want to exercise different parts of the map—exponents, teens, multiples of 10 + other numbers, etc.

tests/nlu/test_tflite_nlu.py
```diff
@@ -0,0 +1,81 @@
...
+
+    model._model._input_details = [{"shape": [1, 50]}]
+    model._model._output_details = [{"shape": [1, 1]}, {"shape": [1, 50, 3]}]
+    model._model.return_value = [np.random.random((1, 1)), np.random.random((1, 50, 3))]
+
+    utterance = "this is only a test"
+    outputs = model(utterance)
+    assert outputs["utterance"] == utterance
+    assert 0.0 <= outputs["confidence"] <= 1.0
+    assert not outputs["slots"]
```
space-pope, line 41: and `outputs["intent"]`?

tests/nlu/test_tflite_nlu.py
```diff
@@ -0,0 +1,81 @@
...
+
+    model._model._input_details = [{"shape": [1, 50]}]
+    model._model._output_details = [{"shape": [1, 1]}, {"shape": [1, 50, 3]}]
+    model._model.return_value = [np.random.random((1, 1)), np.random.random((1, 50, 3))]
+
+    utterance = "this is only a test"
+    outputs = model(utterance)
+    assert outputs["utterance"] == utterance
+    assert 0.0 <= outputs["confidence"] <= 1.0
+    assert outputs["slots"]
```
space-pope, line 81: I think if you add a test for an utterance that contains more than one slot, you'll find a bug in slot parsing. You'll also want to be a bit more specific than `np.random.random` for your outputs—you want to be able to assert that the right intent/slot is picked from a list of options in the metadata.

space-pope (changes requested):
Looks mostly good, but I think there's an issue with slot parsing that needs to be explored with some more tests. Have you run a code coverage tool over this module?

will-rice:
> Looks mostly good, but I think there's an issue with slot parsing that needs to be explored with some more tests. Have you run a code coverage tool over this module?

I'm using coveralls which passed at 100%, but it seems like it is just an indicator that every line is ran at least once. I will port all of the java tests though.

space-pope:
> it is just an indicator that every line is ran at least once

Yeah; that's all coveralls would do, and it can be deceptive. I was missing a couple of the slot tests where things like exponents, etc., are exercised; I thought one or two of the maps might have been left uncovered.

spokestack/nlu/parsers/maps.py
will-rice: I just thought of those when I got to this part. I was trying to think of common variations, but I don't know if they are possible ASR results. My opinion is that it should be optional to use ASR, so we would want common ASR results and possible chat inputs.

space-pope: I agree, but spelling correction for chat is a much bigger project. You have to deal with character insertions/deletions/transpositions, etc. I don't think these are hurting anything, so they can stay, but in general we shouldn't spend too much effort on explicitly including typos in places like this unless they've been observed frequently, and including them gives us some performance benefit.

will-rice: That's a good point. I will remove what isn't included in the mobile libraries.

tests/nlu/parsers/test_digits.py
will-rice: On the metadata violation, I originally sliced the count to the specified length. Is the desired behavior here to not return anything if the metadata is violated?

space-pope: Yes; values that don't match the facets are discarded by the parser. I didn't realize this mismatch the first time through, but your slot parsing doesn't match the API of the other platforms, which ensure that the application has access to the:
* slot name
* parsed value
* raw value recognized in the utterance
In the case of a facet violation, the parsed value would be null, but the raw value would still be present.

spokestack/nlu/tflite.py
```diff
@@ -51,19 +56,39 @@ def __call__(self, utterance: str) -> Dict[str, Any]:
...
+            slot_map[key] = self._tokenizer.decode(value)
 
         # attempt to resolve tagged tokens into slots and
         # collect the successful ones
-        slot_meta = self._intent_meta[intent]["slots"]
-        parsed_slots = []
-        for meta in slot_meta:
-            parsed_slots.append(self._parse_slots(meta, slots))
-
+        slot_meta = {
```
space-pope, line 77: This dict could be created at load time instead of for every classification.

spokestack/nlu/tflite.py
```diff
@@ -51,19 +56,39 @@ def __call__(self, utterance: str) -> Dict[str, Any]:
...
+            meta.pop("name"): meta for meta in self._intent_meta[intent]["slots"]
+        }
+        parsed_slots = {}
+        for key in slot_meta:
+            parsed = self._parse_slots(slot_meta[key], slot_map[key])
+            parsed_slots.update(
```
space-pope, line 83: Another place where `dict[key] = val` would suffice.

tests/nlu/parsers/test_integer.py
```diff
@@ -27,7 +23,85 @@ def test_integer():
...
+    raw_value = "30"
+    parsed = integer.parse(metadata, raw_value)
+    assert parsed == 30
+
+    raw_value = "44"
+    parsed = integer.parse(metadata, raw_value)
+    assert parsed == 44
+
+
+def test_out_of_range():
```
space-pope, line 84: I think a test over ordinals would be useful too. The [inflect](https://pypi.org/project/inflect/) library could be useful for generating values for these tests; you could just loop over every number or a sample of numbers from `1..n` and run it through the parser in its numeric, English cardinal, and English ordinal form, comparing the result to the input.

tests/nlu/test_tflite_nlu.py
```diff
@@ -75,7 +75,82 @@ def test_classify_with_slots(_mock_tokenizer, _mock_model, fs):
     model._model.return_value = [np.random.random((1, 1)), np.random.random((1, 50, 3))]
 
     utterance = "this is only a test"
+    model._encode = mock.MagicMock(return_value=[[utterance], [0, 1, 2, 3, 4, 0]])
     outputs = model(utterance)
+    print(outputs)
```
space-pope, line 80: debug code?

space-pope (commented):
Looking pretty good. Mostly just doc typos and minor things left. I'm OK deferring the test update(s), but it's something to think about.

spokestack/nlu/parsers/digits.py
```diff
@@ -1,13 +1,18 @@
 """
-This module contains the logic to parse digits from NLU results
+This module contains the parser that converts the string reppresentation
```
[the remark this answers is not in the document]
will-rice: *removes useless spellcheck plugin*

spokestack/nlu/tflite.py
```diff
@@ -35,17 +35,21 @@ def __init__(self, model_dir: str) -> None:
         self._intent_meta = {
             intent.pop("name"): intent for intent in self._metadata["intents"]
         }
+        self._slot_meta = {}
+        for intent in self._intent_meta:
+            for slot in self._intent_meta[intent]["slots"]:
+                self._slot_meta[slot.pop("name")] = slot
```
space-pope, line 41: In theory, you could have a slot in two different intents that shares a name but has a different type. This would probably be bad for the tagger itself, but I don't think we currently forbid it anywhere. If that showed up in a model, indexing the slots by name here would result in unexpected behavior, since only one type would be parsable.

spokestack/nlu/tflite.py
```diff
@@ -59,36 +63,31 @@ def __call__(self, utterance: str) -> Dict[str, Any]:
         # retrieve slots from the tagged postions and decode slots back
         # into original values
         slots = [
-            [token_id, tag[2:]] for token_id, tag in zip(input_ids, tags) if tag != "o"
+            (token_id, tag[2:]) for token_id, tag in zip(input_ids, tags) if tag != "o"
         ]
         slot_map: dict = {}
-        for token, tag in slots:
+        for (token, tag) in slots:
+            print(tag)
```
space-pope, line 70: Debug code

space-pope: And this block could also be simplified to
```
if tag in slot_map:
    slot_map[tag].append(token)
else:
    slot_map[tag] = [token]
```
Not a huge savings; just pointing it out.

space-pope (commented):
Good updates. I'd like to see the new print statement gone; the rest of the comments are just comments. The slot map issue is worth looking at, though. I think it's at least partially a question of model architecture whether we should forbid duplicate names, and that decision should inform whether we allow our client libraries to treat duplicates as valid or not.

will-rice:
I'm thinking forbid duplicated names unless there is a good example where duplicate names leads to less complexity for model configuration. It seems like duplicates would be hard to keep up.

space-pope (approved):
Looks good; thanks for the updates. I can see both sides of the slot name thing, but I'm fine forbidding duplicates.

PR #9, opened 2020-09-19
Addition of TTS

will-rice:
This commit introduces text to speech capability to the python library. The following additions have been made:

- TextToSpeechClient: allows one to use spokestack's custom text to speech.
- TextToSpeechManager: this class manages a tts client and an output target.
- PyAudioSpeakerOutput: simple pyaudio interface to the system default speakers


spokestack/tts/clients/spokestack.py
```diff
@@ -0,0 +1,113 @@
...
+        body = self._build_body(utterance, mode, voice)
+        signature = base64.b64encode(
+            hmac.new(self._key, body.encode("utf-8"), hashlib.sha256).digest()
+        ).decode("utf-8")
+        headers = {
+            "Authorization": f"Spokestack {self._key_id}:{signature}",
+            "Content-Type": "application/json",
+        }
+
+        response = requests.post(self._url, headers=headers, data=body).json()
```
brentspell, line 56: you want to use `stream=True` here and return `response.iter_content`, so that you don't wait until all of the audio has been synthesized and downloaded before you return to the client
this also means that the typespec would change

spokestack/tts/clients/spokestack.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if "errors" in response:
+            raise TTSError(response["errors"])
+
+        response = requests.get(response["data"][MODES[mode]]["url"])
```
brentspell, line 61: there is no error checking on this request

spokestack/tts/clients/spokestack.py
brentspell: this code is correctly checking for graphql errors, but you'll also want to check the response for http errors here (response.status_code == 200)

spokestack/tts/tts_manager.py
```diff
@@ -0,0 +1,35 @@
+"""
```
brentspell, line 1: it would be simpler just to call this module `spokestack.tts.manager` instead of `spokestack.tts.tts_manager`

spokestack/tts/tts_manager.py
```diff
@@ -0,0 +1,35 @@
...
+        Args:
+            client: Text to speech client that returns encoded mp3 audio
+            output: Audio outputs target
+    """
+
+    def __init__(self, client, output) -> None:
+        self._client = client
+        self._output = output
+
+    def synthesize(self, **kwargs) -> None:
```
brentspell, line 21: tts parameters should be explicit here - they can't vary across tts clients, and you want the interface to be discoverable to applications

spokestack/tts/tts_manager.py
```diff
@@ -0,0 +1,35 @@
...
+        """ Synthesizes speech and writes to output
+
+        Args:
+            **kwargs: keyword arguments passed to the respective tts clients
```
brentspell, line 25: the tts parameters should be made explicit in the docstring

brentspell (commented):
looks pretty good so far

spokestack/tts/clients/spokestack.py
```diff
@@ -0,0 +1,113 @@
...
+        url (str):
+    """
+
+    def __init__(self, key_id: str, key_secret: str, url: str) -> None:
+
+        self._key_id = key_id
+        self._key = key_secret.encode("utf-8")
+        self._url = url
+
+    def synthesize_speech(
```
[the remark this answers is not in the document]
will-rice: The reason I went with `synthesize_speech` is because that's what polly and google speech use. There is probably a better way to do that than conform our client to their conventions.

brentspell (approved):
lgtm, nice work here

PR #11, opened 2020-09-22
Documentation Build Fix

will-rice:
This should fix the documentation issue with `spokestack.readthedocs.io`. The problem was with `sphinx-autodoc` overwriting the existing `.rst` files which were already set to use the sphinx autodoc feature correctly. This change in workflow is reflected in the README.


brentspell (approved):
it's a bummer that all this stuff has to be committed every time the docs change, but lgtm

will-rice:
yeah I feel like there has to be a way around that because locally you just need the `.rst` files. I'll look into it, maybe I can specify a custom build on readthedocs.

will-rice:
at some point I want to set up a post merge build as well because I'm doing it manually right now.

brentspell:
we only really need to rebuild the docs whenever we release, so it can be included in that process

will-rice:
oh right. that makes more sense.

PR #12, opened 2020-09-29
Addition of speech pipeline

will-rice:
This change introduces the speech pipeline along with some component
interface changes for compatibility. The following major additions have
been added:
- SpeechPipeline class that manages audio input, vad, wakeword, etc.
- ActivationTimeout a pipeline component that puts a lower and upper bound
on the length of time the pipeline is active.


spokestack/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+        self._is_running = False
+
+    def resume(self) -> None:
+        """ Resumes audio input after a pause """
+        self._is_paused = False
+        self._is_running = True
+
+    def run(self) -> None:
+        """ Runs the pipeline to process speech and cleans up after stop is called """
+        while self._is_running:
```
brentspell, line 76: in the case of a synchronous model, where the application calls run on the main thread pauses the pipeline when it receives a recognition event,
`self._is_running` will be set to false when `pause` is called, which will cause the application to exit
it seems that you wouldn't want to set `self._is_running` to false when the pipeline is paused

spokestack/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+            self.step()
+        self.cleanup()
+
+    def step(self) -> None:
+        """ Process a single frame with the pipeline """
+        self._context.event("step")
```
brentspell, line 82: nice, I didn't think about this event, but it could be useful

spokestack/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+        """ Resets all attributes to default configuration """
+        for stage in self._stages:
+            stage.close()
+
+        self._stages.clear()
+        self._input_source.close()
+        self._input_source = None
+        self._context.reset()
+
+    def event(self, function=None, name=None):
```
brentspell, line 95: I thought about this a bit more, and I think `function` should be the final parameter here, so that you could call the decorator without passing the `name` parameter by name:
```
@pipeline.event("recognize")
def handle_recognize(context):
    ....
```

tests/test_activation_timeout.py
```diff
@@ -0,0 +1,19 @@
...
+def test_timeout():
+    max_active = 500
+    context = SpeechContext()
+    timeout = ActivationTimeout(max_active=max_active)
+    context.is_active = True
+
+    for i in range(max_active + 1):
+        timeout(context)
+
+    assert not context.is_active
```
brentspell, line 17: there are more cases that should be tested here - check the logic in the component's `__call__` function and verify that all of the conditions are being exercised

brentspell (commented):
looking pretty good so far, nice work

spokestack/asr/speech_recognizer.py
```diff
@@ -80,7 +80,9 @@ def _receive(self, context):
             hypothesis = hypotheses[0]
             context.transcript = hypothesis["transcript"]
             context.confidence = hypothesis["confidence"]
-        self._is_final = self._client.response.get("final")
+            context.event("partial_recognize")
+        if self._client.response.get("final"):
+            context.event("recognize")
```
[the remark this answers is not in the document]
space-pope: Good thing I read my GitHub emails...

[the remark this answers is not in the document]
brentspell: I think we talked about this in slack a while back, so I figured a fix was in progress
I only mention it here because will's port was correct, but the original was wrong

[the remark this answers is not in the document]
brentspell: https://spokestack.slack.com/archives/CHHAM1519/p1600981282041800

spokestack/context.py
```diff
@@ -77,7 +99,25 @@ def is_active(self, value: bool) -> None:
         Args:
             value (bool): Boolean to set context activity
         """
-        self._active = value
+        self._is_active = value
+
+    @property
+    def is_managed(self) -> bool:
```
space-pope, line 105: I haven't read the entirety of this PR, but I was skimming through and noticed this. This property is something I had to shoehorn in to deal with sharing the microphone with Android platform components; it might not be necessary at all in Python.

will-rice: thank you! yeah we decided to remove the managed stuff so I will make sure this is removed also.

spokestack/asr/speech_recognizer.py
```diff
@@ -80,7 +80,16 @@ def _receive(self, context):
             hypothesis = hypotheses[0]
             context.transcript = hypothesis["transcript"]
             context.confidence = hypothesis["confidence"]
-        self._is_final = self._client.response.get("final")
+            if context.transcript:
+                context.event("partial_recognize")
+            else:
+                context.event("timeout")
```
brentspell, line 86: timeout events should only be raised if we are actually timing out (we have received an empty final transcript), so you don't want to do it for intermediate responses

brentspell (commented):
looks pretty good, just need to fix that one timeout thing

PR #14, opened 2020-10-06
Added ASR API URL as default argument

will-rice:
This change makes the spokestack ASR URL the default argument to the cloud client which mirrors the TTS configuration that came up during feedback.


brentspell (approved):
lgtm

PR #16, opened 2020-10-07
Added Result Class for NLU Results

will-rice:
This change introduces the Result class for catching results from the NLU model. In addition, `typing.py` is added for library type aliases.

### DIFF
--- a/spokestack/nlu/result.py
+++ b/spokestack/nlu/result.py
@@ -1,0 +1,56 @@
+"""
+This module contains the Result class for the NLU.
+"""
+from typing import Any, Dict
+
+from spokestack.typing import Intent, Slot, Utterance
+
+
+class Result:
+    """ Convenience wrapper for NLU Results"""
+
+    def __init__(self):
+        self._utterance: Utterance = None
+        self._intent: Intent = None
+        self._confidence: float = 0.0
+        self._slots: Dict[str, Slot] = {}
+
+    @property
+    def utterance(self) -> Utterance:
+        """ Transcript passed to the NLU model """
+        return self._utterance
+
+    @utterance.setter
+    def utterance(self, value: Utterance) -> None:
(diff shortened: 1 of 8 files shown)


spokestack/nlu/result.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+from typing import Any, Dict
+
+from spokestack.typing import Intent, Slot, Utterance
+
+
+class Result:
+    """ Convenience wrapper for NLU Results"""
+
+    def __init__(self):
```
brentspell, line 12: I think it would be nice to construct the result and pass all of its fields at once

that way, you can make the class immutable (no writable properties) and you can remove the `Union` on all of the type specs, so that the client doesn't have to worry about `None` values

spokestack/typing.py
```diff
@@ -0,0 +1,9 @@
+"""
+This module contains type aliases for the library
+"""
+from typing import Any, Dict, Union
+
+
+Utterance = Union[str, None]
+Intent = Union[str, None]
+Slot = Union[Dict[Any, Any], None]
```
brentspell, line 9: I think we can assume that slot keys will always be `str`

spokestack/nlu/result.py
```diff
@@ -0,0 +1,34 @@
+"""
+This module contains the Result class for the NLU.
+"""
+from typing import Any, Dict
+
+
+class Result:
+    """ Convenience wrapper for NLU Results"""
+
+    def __init__(self, **kwargs) -> None:
```
brentspell, line 10: I would pass these parameters explicitly for documentation and typing

will-rice: ok I believe I have this the right way now. I also added documentation for each argument.

brentspell (approved):
lgtm

PR #17, opened 2020-10-07
Added VAD Trigger

will-rice:
The VAD Trigger allows the speech context to be set to active based on if the vad detects speech.


brentspell (approved):
simple enough, nice work

PR #18, opened 2020-10-08
Install transitive dependencies with spokestack

will-rice:


brentspell (approved):
lgtm, we may want to prune this later, in order to minimize the number of packages that have to be installed for basic use, but I don't think anything here is problematic yet

PR #19, opened 2020-10-09
Addition of module-level logging.

will-rice:
This PR adds logging to the library in similar locations as the android implementation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setuptools.setup(
         "pyaudio",
         "webrtcvad",
         "numpy",
-        "websocket",
+        "websocket_client",
         "tokenizers",
         "requests",
         "streamp3",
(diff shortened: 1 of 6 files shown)


spokestack/wakeword/tflite.py
```diff
@@ -178,6 +184,7 @@ def _detect(self, context: SpeechContext) -> None:
 
         if posterior > self._posterior_threshold:
             context.is_active = True
+            _LOG.info(f"wake: {self._posterior_max}")
```
brentspell, line 187: we want to log the maximum posterior after we update it (I believe this was also wrong in android), so let's move lines 188-189 before 185

setup.py
```diff
@@ -24,7 +24,7 @@
         "pyaudio",
         "webrtcvad",
         "numpy",
-        "websocket",
+        "websocket_client",
```
brentspell, line 27: this should at least be a different commit, probably a separate pr

will-rice: yeah this was an accident. we will make the change though because `websocket_client` is the one we use. there are a couple with similar names which is annoying.

PR #22, opened 2020-10-12
Addition of Pipeline Profiles

will-rice:
This changes introduces two profiles to the python library. Both profiles contain the speech pipeline including asr. The difference is in the activation trigger. WakewordASR uses the tflite wakeword for activation and VoiceActivityTriggerASR uses the vad to activate.


spokestack/profile/vad_trigger_asr.py
```diff
@@ -0,0 +1,59 @@
...
+                    spokestack_id=spokestack_id,
+                    spokestack_secret=spokestack_secret,
+                    sample_rate=sample_rate,
+                    frame_width=frame_width,
+                    **kwargs
+                ),
+            ],
+        )
+
+    def run(self) -> None:
```
brentspell, line 54: the problem with using the profile to hide the pipeline is that you lose all the other features of the pipeline (events, etc.)
it would be better to have the profile construct and return the pipeline, using a `create` method or whatever

spokestack/profile/wakeword_asr.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+from spokestack.activation_timeout import ActivationTimeout
+from spokestack.asr.speech_recognizer import CloudSpeechRecognizer
+from spokestack.io.pyaudio import PyAudioInput
+from spokestack.pipeline import SpeechPipeline
+from spokestack.vad.webrtc import VoiceActivityDetector
+from spokestack.wakeword.tflite import WakewordTrigger
+
+
+class WakewordASR:
```
brentspell, line 12: asr here would probably be better qualified as the spokestack asr, so that you can later add other asr profiles

spokestack/profile/vad_trigger_asr.py
```diff
@@ -0,0 +1,59 @@
...
+Pipeline profile with vad trigger and asr
+"""
+from spokestack.activation_timeout import ActivationTimeout
+from spokestack.asr.speech_recognizer import CloudSpeechRecognizer
+from spokestack.io.pyaudio import PyAudioInput
+from spokestack.pipeline import SpeechPipeline
+from spokestack.vad.webrtc import VoiceActivityDetector, VoiceActivityTrigger
+
+
+class VoiceActivityTriggerASR:
```
brentspell, line 11: asr here would probably be better qualified as the spokestack asr, so that you can later add other asr profiles

spokestack/profile/vad_trigger_asr.py
will-rice: one thought I had originally was being able to add the event handlers like `@profile.pipeline.event`

spokestack/profile/vad_trigger_asr.py
```diff
@@ -0,0 +1,59 @@
...
+    """ VAD Trigger ASR
+
+    Args:
+            spokestack_id (str):
+            spokestack_secret (str):
+            sample_rate (int):
```
will-rice, line 17: forgot the doc strings...

spokestack/profile/vad_trigger_asr.py
brentspell: I think that the primary thing the application interacts with is the pipeline, whereas the profile just provides some boilerplate configuration and wiring

spokestack/profile/vad_trigger_asr.py
```diff
@@ -0,0 +1,53 @@
...
+from spokestack.asr.speech_recognizer import CloudSpeechRecognizer
+from spokestack.io.pyaudio import PyAudioInput
+from spokestack.pipeline import SpeechPipeline
+from spokestack.vad.webrtc import VoiceActivityDetector, VoiceActivityTrigger
+
+
+class VoiceActivityTriggerSpokestackASR:
+    """ VAD Trigger ASR """
+
+    def create(
```
brentspell, line 14: nice, since this class doesn't have any state, you could mark this method with `@staticmethod` and just call it like `VoiceActivityTriggerSpokestackASR.create()` without creating an instance

will-rice: TIL. thanks!!

brentspell (approved):
lgtm

PR #23, opened 2020-10-13
Refactor of TTS Return Value

will-rice:
This change adds the ability to synthesize text and get the audio via url.


spokestack/io/pyaudio.py
```diff
@@ -3,6 +3,10 @@
 """
 import numpy as np  # type: ignore
 import pyaudio  # type: ignore
+from streamp3 import MP3Decoder  # type: ignore
```
brentspell, line 6: we don't want mp3 or tts details in the pyaudio stuff
this module should just be a simple abstraction around pyaudio

spokestack/tts/clients/spokestack.py
```diff
@@ -65,12 +73,14 @@ def synthesize(
         if "errors" in response:
             raise TTSError(response["errors"])
 
-        response = requests.get(response["data"][_MODES[mode]]["url"], stream=True)
+        response = requests.get(
```
brentspell, line 76: this http request should not be made if you are only requesting the audio url
otherwise, the http request will be made twice - once by spokestack and then again by the application
this is why url-based tts should probably just be a separate method

spokestack/tts/clients/spokestack.py
```diff
@@ -72,6 +72,38 @@ def synthesize(
 
         return response.iter_content(chunk_size=None)
 
+    def synthesize_url(
```
brentspell, line 75: we can remove all this duplication by having `synthesize` call `synthesize_url` and use the result to download the audio

will-rice: yep should have been able to see that one. thanks

brentspell: a lot of times it takes a second pair of eyes

brentspell (approved):
lgtm

PR #25, opened 2020-10-22
Addition of Google Cloud as an ASR provider.

will-rice:


.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ jobs:
     build:
         working_directory: ~/repo
         docker:
-            - image: circleci/python:3.7
```
brentspell, line 7: this is another set of changes that should be on their own commit
this would be a good opportunity for you to learn to rewrite your changes as separate commits, so that the commit that adds google cloud as an asr provider doesn't have any of this stuff in it

spokestack/asr/spokestack/cloud_client.py
```diff
@@ -13,17 +13,17 @@
 
 
 class CloudClient:
-    """ Spokestack client for cloud based speech to text
```
brentspell, line 16: this rename will be a breaking change, so it should be deployed as a new major release number, and you'll want to check all documentation/blog posts/example projects and upgrade them
also, these changes should be on a separate commit, which is also separate from the formatter changes

spokestack/asr/spokestack/cloud_client.py
```diff
@@ -13,17 +13,17 @@
...
+    """Spokestack client for cloud based speech to text
 
-        Args:
```
brentspell, line 18: for these mass changes to the formatter, we will want to create a .git-blame-ignore-revs file so that they are ignored when viewing the history
https://akrabat.com/ignoring-revisions-with-git-blame/

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+_LOG = logging.getLogger(__name__)
+
+
+class GoogleSpeechRecognizer:
+    """ Transforms speech into text using Google's ASR.
+
+     Args:
+         cred_path (str): path to google credentials
```
brentspell, line 23: we will also want to support credentials as strings (environment variables, etc.), so that we don't have to put them into files
you can probably deal with this with separate kwargs

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,104 @@
...
+         language (str): language of input audio
+         sample_rate (int): sample rate of the input audio (Hz)
+         **kwargs (optional): additional keyword arguments
+     """
+
+    def __init__(
+        self, cred_path, language="en-US", sample_rate=16000, **kwargs,
```
brentspell, line 30: we don't want language defaults biased to en-US, this parameter should always be required

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _receive(self, context):
+
+        for response in self._client.streaming_recognize(self._config, self._drain()):
+            for result in response.results[:1]:
+                for alternative in result.alternatives[:1]:
+                    context.transcript = alternative.transcript
+                    context.confidence = alternative.confidence
+
+                if result.is_final:
```
brentspell, line 78: we should also support partial result events - see how this is done in the spokestack asr client

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,104 @@
...
+        self._queue.put(None)
+        self._thread.join()
+        self._thread = None
+
+    def _send(self, frame) -> None:
+        self._queue.put(speech.StreamingRecognizeRequest(audio_content=frame.tobytes()))
+
+    def reset(self) -> None:
+        """ resets recognizer """
+        self._queue.empty()
```
brentspell, line 97: you'll need to send a `None` value here, or you'll deadlock the thread when you call join

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,104 @@
...
+        credentials = service_account.Credentials.from_service_account_file(cred_path)
+        self._client = speech.SpeechClient(credentials=credentials, **kwargs)
+        self._config = speech.StreamingRecognitionConfig(
+            config=speech.RecognitionConfig(
+                encoding=speech.RecognitionConfig.AudioEncoding.LINEAR16,
+                sample_rate_hertz=sample_rate,
+                language_code=language,
+                enable_automatic_punctuation=True,
+            ),
+            interim_results=False,
```
brentspell, line 42: we'll want interim_results = True for our partial result events

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,104 @@
...
+         sample_rate (int): sample rate of the input audio (Hz)
+         **kwargs (optional): additional keyword arguments
+     """
+
+    def __init__(
+        self, cred_path, language="en-US", sample_rate=16000, **kwargs,
+    ) -> None:
+
+        credentials = service_account.Credentials.from_service_account_file(cred_path)
+        self._client = speech.SpeechClient(credentials=credentials, **kwargs)
```
brentspell, line 34: we don't want to forward spokestack kwargs to the speech client, because we will be sending it things like vad configuration, etc.
I'm guessing this is just working by accident right now

spokestack/asr/spokestack/cloud_client.py
will-rice: can we do this with the documentation html stuff also?

spokestack/asr/google/speech_recognizer.py
will-rice: I thought the point of kwargs was that you could add arbitrary keyword arguments as long as each class has `**kwargs` in its signature.

spokestack/asr/spokestack/cloud_client.py
will-rice: I didn't think about the post changes. Would it be better to allow this to be imported from any level under `asr`? I think that would allow all of the old imports to work, but going forward each bottom class would need a unique name.

brentspell: it's not a big deal that we're changing/breaking things, we just need to make sure to do all the right things with versioning and documentation
it should be one way or the other, though, we don't want multiple ways to import the same thing

spokestack/asr/spokestack/cloud_client.py
brentspell: it doesn't matter for generated things like documentation
we just want to preserve readable diffs/blames from sweeping changes for things that are maintained/viewed manually

spokestack/asr/google/speech_recognizer.py
brentspell: right, but we don't want to make assumptions about the speech client's parameters or accidentally pass a conflicting parameter to them

spokestack/asr/spokestack/cloud_client.py
brentspell: that said, we would still want the generated documentation stuff on its own commits, probably in the same commit that updates the release number

will-rice: ok that makes sense. I'll do my feature commit and then another one with the docs build and version change.

spokestack/asr/spokestack/cloud_client.py
will-rice: I hate coming across an outdated post/repo so I agree updating is the best idea. for all of the posts we can just find and replace. At first that seemed like a bigger problem than it actually is. also having multiple aliases like we've seen in tensorflow can be really frustrating when one of them is deprecated.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# black format change
+515508d279c8be35ce928ea4dd36f16484665e04
```
brentspell: we'll want to remove this change

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+_LOG = logging.getLogger(__name__)
+
+
+class GoogleSpeechRecognizer:
+    """Transforms speech into text using Google's ASR.
+
+    Args:
+        language (str): language of input audio
```
brentspell, line 23: duplicate docstring line, also might be good to document the language code format and add examples

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+        credentials (optional): google API credentials or path to credentials.
+                                if set to None these will be pulled from the
+                                environment variable: GOOGLE_APPLICATION_CREDENTIALS
+        sample_rate (int): sample rate of the input audio (Hz)
+        **kwargs (optional): additional keyword arguments
+    """
+
+    def __init__(
+        self, language, credentials=None, sample_rate=16000, **kwargs,
```
brentspell, line 33: type hints would be good here, especially since credentials can be `None`, `str`, or `dict`

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+    ) -> None:
+        if credentials:
+            if isinstance(credentials, str):
+                credentials = service_account.Credentials.from_service_account_file(
+                    credentials
+                )
+            else:
```
brentspell, line 40: might be good to add an explicit check for `dict` here and raise a valueerror on bad credentials so the user will know what is going on

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+                for alternative in result.alternatives[:1]:
+                    context.transcript = alternative.transcript
+                    context.confidence = alternative.confidence
+
+                if result.is_final:
+                    if context.transcript:
+                        context.event("recognize")
+                        _LOG.debug("recognize event")
+                else:
+                    context.event("partial_recognize")
```
brentspell, line 94: the spokestack recognizer doesn't raise the partial event if the transcript is empty, this recognizer should be consistent with that

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+    def _receive(self, context):
+
+        for response in self._client.streaming_recognize(self._config, self._drain()):
+            for result in response.results[:1]:
+                for alternative in result.alternatives[:1]:
+                    context.transcript = alternative.transcript
+                    context.confidence = alternative.confidence
+
+                if result.is_final:
+                    if context.transcript:
```
brentspell, line 90: the spokestack recognizer raises the timeout event on the final transcript if it is empty, this recognizer should be consistent with that

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+                        context.event("recognize")
+                        _LOG.debug("recognize event")
+                else:
+                    context.event("partial_recognize")
+                    _LOG.debug("partial recognize event")
```
brentspell, line 95: the spokestack recognizer doesn't log partial events, this should be made consistent

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+_LOG = logging.getLogger(__name__)
+
+
+class GoogleSpeechRecognizer:
+    """Transforms speech into text using Google's ASR.
+
+    Args:
+        language (str): language of input audio
+        language (str): language of input audio
+        credentials (optional): google API credentials or path to credentials.
```
brentspell, line 25: it would be good to indicate that hard-coded credentials are expressed as a dict

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+        self._config = speech.StreamingRecognitionConfig(
+            config=speech.RecognitionConfig(
+                encoding=speech.RecognitionConfig.AudioEncoding.LINEAR16,
+                sample_rate_hertz=sample_rate,
+                language_code=language,
+                enable_automatic_punctuation=True,
+            ),
+            interim_results=True,
+        )
+        self._sample_rate = sample_rate
```
brentspell, line 55: `_sample_rate` and `_language` are not used

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+        self._queue.put(None)
+        self._thread.join()
+        self._thread = None
+
+    def _send(self, frame) -> None:
+        self._queue.put(speech.StreamingRecognizeRequest(audio_content=frame.tobytes()))
+
+    def reset(self) -> None:
+        """ resets recognizer """
+        self._queue.put(None)
```
brentspell, line 111: adding a value to the queue and then emptying it does nothing
also, there is no need to write a value to the queue if the thread is not attached

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,119 @@
...
+            self._commit()
+        if context.is_active:
+            self._send(frame)
+
+    def _begin(self, context) -> None:
+        self._thread = Thread(target=self._receive, args=(context,),)
+        self._thread.start()
+
+    def _receive(self, context):
+
```
brentspell, line 82: this blank line is inconsistent with the rest of the class's methods

brentspell (commented):
pretty close here, just a few things to correct

spokestack/asr/google/speech_recognizer.py
```diff
@@ -0,0 +1,128 @@
...
+                                              environment variable:
+                                              GOOGLE_APPLICATION_CREDENTIALS
+        sample_rate (int): sample rate of the input audio (Hz)
+        **kwargs (optional): additional keyword arguments
+    """
+
+    def __init__(
+        self,
+        language: str,
+        credentials=Union[None, str, dict],
```
brentspell, line 38: is this syntax correct?
seems like it should be `credentials: Union...`

brentspell: and it should probably default to `None`

will-rice: I fixed it.

brentspell (approved):

PR #29, opened 2020-11-16
Organization and Content update to the README

will-rice:
This change adds content to the README for initializing components of the library while briefly explaining each component.


README.md
```diff
@@ -1,71 +1,214 @@
...
+
+#### Tensorflow
+
+The full Tensorflow package is installed with the following:
+
+```shell
+pip install tensorflow
+```
+
+#### TFLite Iterpreter (Embedded Devices)
```
will-rice, line 79: "Interpreter" I'll fix this.

brentspell:
might be better to get josh to take a look at this one for consistency across the platforms

will-rice:
@space-pope just wanted to point out that this is more of an introduction. There will be actual docs added to the site

README.md
```diff
@@ -1,71 +1,214 @@
...
+### Installation with pip
 
-    pip install spokestack
+Once system dependencies have been satisfied, you can install the library with the following.
+
+```shell
+pip install spokestack
+```
 
 ## Development
```
space-pope, line 54: Typically, a "development" section in a library readme is for people contributing to the library, not using it.

space-pope (changes requested):
Your comment touched on my main feedback—a lot of what's here would be better suited for site documentation.

I mentioned this in one of the comments, but my main advice is to flip what you've got here upside down: introduce the full working speech pipeline with a self-contained working snippet, then if necessary go into details below. Some mentions of the individual components can stay here for now, but I think eventually you'll just want to turn most of that stuff into links to site documentation. 

IMO, a readme should be an introduction to what the library does, the briefest of quickstarts, and any helpful troubleshooting tips that aren't immediately obvious. That changes a bit if you don't have an API reference or explanatory documentation elsewhere, but we do, so I lean toward our readmes being more of a foot in the door that gives the basics but points people to the site for more info.

Reading this also reminded me that the other platforms have recently added a `Spokestack` wrapper that sets up the speech pipeline, nlu, and tts all at once. None of the old stuff has gone away, but it's easier to use them all together now. Something to think about for platform parity with Python.

README.md
```diff
@@ -1,71 +1,214 @@
...
+
+This library requires a way to run [TFLite](https://www.tensorflow.org/lite) models. There are two primary ways to add this ability. The first, is installing the full [Tensorflow](https://www.tensorflow.org/) library. In most instances you should be able to install this with `pip`, but in the case of any installation issues follow the installation [guide](https://www.tensorflow.org/install).
+
+#### Tensorflow
+
+The full Tensorflow package is installed with the following:
+
+```shell
+pip install tensorflow
+```
```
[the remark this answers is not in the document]
will-rice: I think you are better off covering that the library requires a way to run tensorflow models than specifying which parts need tensorflow. Saying tensorflow is required needs to be written one time, but specifying the components will change over time.

README.md
```diff
@@ -1,71 +1,214 @@
...
+
+```python
+from spokestack.vad.webrtc import VoiceActivityTrigger
+
+vad = VoiceActivityTrigger()
+```
+
+### Wakeword Detection
+
+Also known as keyword spotting (kws), wakeword is the way to get your applications attention with voice. This is normally through a specific keyword or phrase. The default wakeword for this library is "Spokestack". Wakeword can be added like this:
```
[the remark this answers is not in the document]
will-rice: It's the only model included in a spokestack account by default. it's the wakeword for all of the tutorials. it's the default wakeword. This seems like an arbitrary distinction.

README.md
```diff
@@ -1,71 +1,214 @@
...
+
+Automatic Speech Recognition is a technique used to turn speech into a transcript. You can initialize our cloud ASR with:
+
+```python
+from spokestack.asr.spokestack.speech_recognizer import SpeechRecognizer
+
+asr = SpeechRecognizer("spokestack_id", "spokestack_secret")
+```
+
+### Speech Pipeline
```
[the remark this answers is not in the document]
will-rice: that makes sense. I really didn't know how to arrange it so thanks. I will remember this.

README.md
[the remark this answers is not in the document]
space-pope: It's a bias I have from focusing on modularity in the Android library—if people only want to use ASR and TTS, they don't need TF. This is maybe less of a concern in Python, where there's no high-quality "on-device" ASR that comes with the platform by default.

README.md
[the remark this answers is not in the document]
space-pope: The difference is that the library provides a way to _run_ wakeword detection, not a wakeword itself. The library is code; the wakeword models are data. We give out one wakeword for free, so it's the one we use in the tutorials, but we don't distribute it with the libraries themselves, so the libraries don't _have_ a wakeword by default (you can't use a wakeword in your app without a separate download); they just have a component capable of using the models.

Like I said, it's a nitpick as far as readme wording goes (once you add the links to the wakeword models so that people don't assume it works out of the box); I'm just usually biased toward precision, sometimes overly so.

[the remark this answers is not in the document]
will-rice: It might be better to make it work out of the box.

README.md
will-rice: what should I call this instead?

will-rice: I ended up calling it. "Using the Library"

README.md
```diff
@@ -51,7 +51,7 @@ Once system dependencies have been satisfied, you can install the library with t
 pip install spokestack
 ```
 
-## Development
+## Using the Library
```
space-pope, line 54: Yeah, that works. I'd personally go with "usage" (shorter/consistent with other platforms), but this means the same.

README.md
```diff
@@ -77,6 +77,36 @@ pip install tensorflow
 
 In use cases where you require a small footprint, such as on a raspberry pi or similar embedded devices, you will want to install the TFLite Interpreter. You can install it for your platform by following [the instructions here](https://www.tensorflow.org/lite/guide/python#install_just_the_tensorflow_lite_interpreter).
 
+### Speech Pipeline
+
+The Speech Pipeline is the component that ties together VAD, Wakeword, and ASR. Essentially, this component listens to a frame of audio to determine if speech is present. If speech is detected, the Wakeword model processes the subsequent frames of audio looking for the specific keyword. If the keyword is found, the pipeline is activated and converts the following audio into a transcript. The Speech Pipeline is established like this from the previous initialized components:
```
space-pope, line 82:
```suggestion
The Speech Pipeline is the module that ties together VAD, Wakeword, and ASR. The VAD (voice activity detection) listens to a frame of audio captured by the input device to determine if speech is present. If it is, the Wakeword model processes subsequent frames of audio looking for its keyword. If the keyword is found, the pipeline is activated and converts the following audio into a transcript. The Speech Pipeline is initialized like this:
```

Edited slightly for consistency with the terminology in the other platforms and because changing the order of the paragraphs changes what needs to be defined/introduced a bit.

README.md
```diff
@@ -77,6 +77,36 @@ pip install tensorflow
...
+
+Pipeline callbacks allow additional code to be executed when a speech event is detected. For example, we can print when the pipeline is activated by registering a function with the `pipeline.event` decorator.
+
+```python
+@pipeline.event
+def on_activate(context):
+    print(context.is_active)
+```
+
 ### Audio Input
```
space-pope, line 110: Now that the structure has been flipped, there's actually little need to go over initialization of the individual components since it's all in the first section. You could have a "Pipeline components" section where you give some more details on each component, or you could get rid of the individual component explanations altogether in favor of some more detail on pipeline callbacks, which are what people need to know about in order to get the ASR transcripts. Right now, there are instructions on how to set up the pipeline, but not too much on how to make use of it.

If you do keep the sections on individual components, they should at least be demoted a header level or two so they're not as prominent as the pipeline, NLU, and TTS.

README.md
```diff
@@ -111,49 +141,17 @@ wake = WakewordDetector("path_to_tflite_model")
...
-```python
-@pipeline.event
-def on_activate(context):
-    print(context.is_active)
-```
-
 ### Natural Language Understanding (NLU)
 
-Natural Language Understanding turns an utterance into a machine readable format. For our purposes, this is joint intent detection and slot filling. We like to think of intents as the action a user desires from an application, and slots as the optional arguments to fulfill the requested action. Our NLU model is initialized like this:
+Natural Language Understanding turns an utterance into a machine readable format. For our purposes, this is joint intent detection and slot filling. You can read more about the concepts [here](https://www.spokestack.io/docs/concepts/nlu). We like to think of intents as the action a user desires from an application, and slots as the optional arguments to fulfill the requested action. Our NLU model is initialized like this:
```
space-pope, line 154: How does this fit in with the pipeline introduced above?

space-pope (commented):
Looking better; thanks for the update. Think we just need a bit more to tie the usage section together (and maybe we can drop some of the component documentation stubs), and we should be good to go.

README.md
will-rice: I guess it doesn't without the Spokestack abstraction.

space-pope: Oh, I wasn't suggesting that it _doesn't_ fit, just that a snippet that uses the NLU in a pipeline transcript callback would be helpful to the reader.

The Spokestack abstraction is another layer of developer ergonomics, but the library's totally usable without it, especially in Python where initialization of the separate components doesn't take up a screen of code all by itself.

will-rice: so I guess what would be good for the readme is like an abridged tutorial. overview of all the features, but super brief.

space-pope: That's what I find most useful in a readme. As @noelweichbrodt has pointed out, this holds the potential for getting out of sync with the codebase, so sticking to the highest conceptual level possible is best, but I still think it's good to give just enough depth that the readme can provide a working, paste-able example of the library's main value proposition(s).

Link to the docs for more details, but provide enough value in the readme that the reader _wants_ to make that extra click.

will-rice: ok I think I understand. sorry this was such a big deal.

space-pope: No problem at all—I hate to keep asking for changes, but I have...strong opinions about words (and like to present as unified a front as possible across the platforms).

README.md
```diff
@@ -1,71 +1,196 @@
...
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
-[![Documentation Status](https://readthedocs.org/projects/spokestack/badge/?version=latest)](https://spokestack.readthedocs.io/en/latest/?badge=latest)
+[![GitHub stars](https://img.shields.io/github/stars/spokestack/spokestack-python?style=social)](https://github.com/spokestack/spokestack-python/stargazers)
+![GitHub watchers](https://img.shields.io/github/watchers/spokestack/spokestack-python?style=social)
+![Twitter Follow](https://img.shields.io/twitter/follow/spokestack?style=social)
+
+Welcome to Spokestack Python! This library is intended for developing [voice interfaces](https://en.wikipedia.org/wiki/Voice_user_interface) in Python. This can include anything from [Raspberry Pi](https://www.raspberrypi.org/) applications like traditional smart speakers to [Django](https://www.djangoproject.com/) web applications. Literally anything built in [Python](https://www.python.org/) can be given a voice interface.
```
noelweichbrodt, line 16:
```suggestion
Welcome to Spokestack Python! This library is intended for developing [voice interfaces](https://en.wikipedia.org/wiki/Voice_user_interface) in Python. This can include anything from [Raspberry Pi](https://www.raspberrypi.org/) applications like traditional smart speakers to [Django](https://www.djangoproject.com/) web applications. *Anything* built in [Python](https://www.python.org/) can be given a voice interface.
```

README.md
```diff
@@ -1,71 +1,196 @@
...
+
+The full Tensorflow package is installed with the following:
+
+```shell
+pip install tensorflow
+```
+
+#### TFLite Interpreter (Embedded Devices)
+
+In use cases where you require a small footprint, such as on a raspberry pi or similar embedded devices, you will want to install the TFLite Interpreter. You can install it for your platform by following [the instructions here](https://www.tensorflow.org/lite/guide/python#install_just_the_tensorflow_lite_interpreter).
```
noelweichbrodt, line 79:
```suggestion
In use cases where you require a small footprint, such as on a Raspberry pi or similar embedded devices, you will want to install the TFLite Interpreter. You can install it for your platform by [following the instructions](https://www.tensorflow.org/lite/guide/python#install_just_the_tensorflow_lite_interpreter).
```

README.md
```diff
@@ -1,71 +1,196 @@
...
+
+### Speech Pipeline
+
+The Speech Pipeline is the module that ties together VAD, Wakeword, and ASR. The VAD (voice activity detection) listens to a frame of audio captured by the input device to determine if speech is present. If it is, the Wakeword model processes subsequent frames of audio looking for its keyword. If the keyword is found, the pipeline is activated and converts the following audio into a transcript. The Speech Pipeline is initialized like this:
```
noelweichbrodt, line 83:
```suggestion
The Speech Pipeline is the module that ties together VAD (voice activity detection), wakeword, and ASR (automated speech detection). The VAD listens to a frame of audio captured by the input device to determine if speech is present. If it is, the wakeword model processes subsequent frames of audio looking for the keyword it has been trained to recognize. If the keyword is found, the pipeline is activated and performs speech recognition, converting the subsequent audio into a transcript. The Speech Pipeline is initialized like this:
```

noelweichbrodt (commented):
This is good stuff! After a cycle or two of updating both the readme and the docs site you will find it easier to transition much of this into the docs site and use the readme to point people to various docs.

README.md
```diff
@@ -1,71 +1,196 @@
...
-## Development
+Once system dependencies have been satisfied, you can install the library with the following.
+
+```shell
+pip install spokestack
+```
+
+## Usage
 
 ### Setup
```
space-pope, line 56: The setup and TF sections feel more like part of `Get Started` rather than `Usage` since we're still talking about dependencies.

README.md
```diff
@@ -1,71 +1,196 @@
...
+
+The full Tensorflow package is installed with the following:
+
+```shell
+pip install tensorflow
+```
+
+#### TFLite Interpreter (Embedded Devices)
+
+In use cases where you require a small footprint, such as on a Raspberry pi or similar embedded devices, you will want to install the TFLite Interpreter. You can install it for your platform by [following the instructions](https://www.tensorflow.org/lite/guide/python#install_just_the_tensorflow_lite_interpreter).
```
space-pope, line 79:
```suggestion
In use cases where you require a small footprint, such as on a Raspberry Pi or similar embedded devices, you will want to install the TFLite Interpreter. You can install it for your platform by [following the instructions](https://www.tensorflow.org/lite/guide/python#install_just_the_tensorflow_lite_interpreter).
```
Just for consistency with the previous mention.

README.md
```diff
@@ -1,71 +1,196 @@
...
+
+```python
+@pipeline.event
+def on_recognize(context):
+    print(context.transcript)
+```
+
+### Natural Language Understanding (NLU)
+
+Natural Language Understanding turns an utterance into a machine readable format. For our purposes, this is joint intent detection and slot filling. You can read more about the concepts [here](https://www.spokestack.io/docs/concepts/nlu). We like to think of intents as the action a user desires from an application, and slots as the optional arguments to fulfill the requested action. Our NLU model is initialized like this:
```
space-pope, line 125:
```suggestion
Natural Language Understanding turns an utterance into structured data a machine can act on. For our purposes, this is joint intent detection and slot filling. You can read more about the concepts [here](https://www.spokestack.io/docs/concepts/nlu). We like to think of intents as the action a user desires from an application, and slots as the optional arguments to fulfill the requested action. Our NLU model is initialized like this:
```

The fact that it's so hard to define this in a way that makes it sound like a useful addition should be a red flag to the community that maybe their current approach doesn't add that much value...

space-pope (commented):
Just a couple things I noticed from re-reading the whole thing as a unit. Looking good!

space-pope (approved):
Great! Thanks for sticking with it.

PR #31, opened 2020-12-14
Addition of Keyword Recognizer

will-rice:
This PR adds a keyword recognizer class that detects a set of keywords in an audio stream.


spokestack/asr/spokestack/keyword_recognizer.py
```diff
@@ -0,0 +1,195 @@
...
+"""
+import logging
+import os
+from typing import List
+
+import numpy as np  # type: ignore
+
+from spokestack.context import SpeechContext
+from spokestack.models.tensorflow import TFLiteModel
+from spokestack.wakeword.ring_buffer import RingBuffer
```
brentspell, line 13: it would be better to pull the ring buffer module up to the root spokestack package, so you don't have dependences on the wakeword package

spokestack/asr/spokestack/keyword_recognizer.py
```diff
@@ -0,0 +1,195 @@
+"""
```
brentspell, line 1: it's a bit weird to call this package `keyword_recognizer` and call the similar wakeword package `tflite`
if it's really important for the client to know that they are using tflite, then this module should probably be tflite.py under spokestack/asr/keyword/
if the client doesn't care about tflite, then something like spokestack/asr/keyword/recognizer.py might be simpler
personally, I doubt we will ever implement alternatives to the wakeword/keyword detectors we have now

spokestack/asr/spokestack/keyword_recognizer.py
```diff
@@ -0,0 +1,195 @@
...
+        posterior = self.detect_model(frame)[0][0]
+        class_index = np.argmax(posterior, -1)
+        confidence = posterior[class_index]
+        print(confidence)
+
+        if confidence >= self._posterior_threshold:
+            context.transcript = self.classes[class_index]
+            context.confidence = confidence
+            context.event("recognize")
+            _LOG.debug("recognize event")
```
brentspell, line 180: since you're already raising events for these, you probably don't need debug logs that say the same thing, especially since you already have a logging statement above

spokestack/asr/spokestack/keyword_recognizer.py
```diff
@@ -0,0 +1,195 @@
...
+        pre_emphasis: float = 0.97,
+        sample_rate: int = 16000,
+        fft_window_type: str = "hann",
+        fft_hop_length: int = 10,
+        model_dir: str = "",
+        posterior_threshold: float = 0.5,
+        **kwargs
+    ) -> None:
+
+        self.classes = classes
```
brentspell, line 44: it would be good to validate the length of the class list against the output shape of the detector model

spokestack/asr/spokestack/keyword_recognizer.py
```diff
@@ -0,0 +1,195 @@
...
+        self.encode_window.write(frame)
+
+    def _detect(self, context: SpeechContext) -> None:
+        # read the full contents of the encode window and add the batch dimension
+        # calculate a scalar likelihood that the frame contains a keyword
+        # with the detect model
+        frame = self.encode_window.read_all()
+        frame = np.expand_dims(frame, 0)
+        posterior = self.detect_model(frame)[0][0]
+        class_index = np.argmax(posterior, -1)
```
brentspell, line 172: shouldn't need the -1 here, since there's only one dimension

brentspell (commented):
nice work here, just a couple of small improvements

spokestack/asr/spokestack/keyword_recognizer.py
```diff
@@ -0,0 +1,195 @@
...
+        confidence = posterior[class_index]
+        print(confidence)
```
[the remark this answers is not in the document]
will-rice: this shouldn't even be here, it was for debugging. I will find/make a pre-commit to look for print statements.

spokestack/asr/spokestack/keyword_recognizer.py
will-rice: Yeah, I'm not sure if `tflite` is important to the client. Initially, the thought was that we could add support for Onnx/PyTorch. However, in that case, it would be better to not change the top-level interface and just allow a backend swap.

tests/wakeword/test_ring_buffer.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 import pytest
 
-from spokestack.wakeword.ring_buffer import RingBuffer
```
brentspell, line 7: this test should be moved out of the wakeword directory

tests/asr/keyword/test_tflite_recognizer.py
```diff
@@ -0,0 +1,88 @@
+"""
```
brentspell, line 1: the convention for tests is that the module name should be the same as the module under test, with the "test_" prefix

brentspell (approved):
good work here and thanks for the rebase

PR #33, opened 2021-01-18
Addition of Gain Control, Noise Suppression, and VAD

will-rice:
Addtion of Gain Control, Noise Suppression, and VAD.

This commit introduces automatic gain control and automatic noise suppression from the filter_audio (webrtc) library. In addition, the webrtcvad dependency has been removed in favor of using our VAD extension. The major changes are:
- Added AutomaticGainControl extension
- Added AutomaticNoiseSuppression extension
- Added WebRtcVad extension
- Changed webrtcvad to our WebRtcVad extension


brentspell:
First of all, nice work here. I suspect it was quite the learning curve, but it was good to get some experience with cython.

brentspell:
One thing we'll want to do is to add the new components to the pipeline profiles, so that the profile configurations line up with those in the android repo. This can be done on a separate PR if you want.

https://github.com/spokestack/spokestack-android/tree/master/src/main/java/io/spokestack/spokestack/profile

setup.py
```diff
@@ -1,12 +1,68 @@
...
+        "filter_audio/vad/webrtc_vad.c",
+        "filter_audio/agc/analog_agc.c",
+        "filter_audio/agc/digital_agc.c",
+        "filter_audio/ns/nsx_core.c",
+        "filter_audio/ns/nsx_core_c.c",
+        "filter_audio/ns/noise_suppression_x.c",
+    ]
+]
+
+EXTENSIONS = [
```
brentspell, line 43: We have a bit of an inconsistency in naming of the webrtc stuff now from the client's perspective. For each of the webrtc components, this is what the client would have to import.

```
spokestack.extensions.webrtc.agc.AutomaticGainControl
spokestack.extensions.webrtc.ans.AutomaticNoiseSuppression
spokestack.vad.webrtc.VoiceActivityDetector/VoiceActivityTrigger
```

The client doesn't really care about the term `extension`, so I would recommend making new modules/classes `spokestack.agc.webrtc.AutomaticGainControl` and `spokestack.nsx.webrtc.AutomaticNoiseSuppression` if you can get them to build like that. If not, see if you can re-export those classes under these module names instead of the `extension` module names.

setup.py
```diff
@@ -1,12 +1,68 @@
...
+    setuptools.Extension(
+        "spokestack.extensions.webrtc.agc",
+        ["spokestack/extensions/webrtc/agc.pyx"] + SOURCES,
+        include_dirs=["filter_audio/agc/include/"],
+    ),
+    setuptools.Extension(
+        "spokestack.extensions.webrtc.ans",
```
brentspell, line 50: I think a better acronym for for the noise suppressor would be `nsx`, which would be consistent with the terminology in the webrtc library. The android library got this wrong, but at the time, I didn't notice the inconsistency. You have a chance to fix it here.

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+# distutils: sources = filter_audio/agc/analog_agc.c, filter_audio/agc/digital_agc.c
+# distutils: include_dirs = filter_audio/agc/include/
+cimport numpy as np
+cimport cagc
+
+
+MIC_MAX = 255
+MIC_TARGET = 180
+
+np.import_array()
```
brentspell, line 13: Is this statement necessary? I remember removing it in a version I was compiling and everything was fine. If it is needed, disregard this comment.

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+
+cdef class AutomaticGainControl:
+    cdef void*_agc
+    cdef int _sample_rate
```
brentspell, line 17: The only member variable that is actually used outside the constructor is the `_agc` member. The rest can be removed.

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+cimport cagc
+
+
+MIC_MAX = 255
+MIC_TARGET = 180
+
+np.import_array()
+
+cdef class AutomaticGainControl:
+    cdef void*_agc
```
brentspell, line 16: I would add a space here between `void*` and `_agc`

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+    cdef int _frame_width
+    cdef int _target_level_dbfs
+    cdef int _compression_gain_db
+    cdef int _limit_enable
+    cdef cagc.WebRtcAgc_config_t _config
+
+    def __dealloc__(self):
+        cagc.WebRtcAgc_Free(self._agc)
+
+    def __init__(self,
```
brentspell, line 27: This constructor need to validate its parameters like the android version.

https://github.com/spokestack/spokestack-android/blob/master/src/main/java/io/spokestack/spokestack/webrtc/AutomaticGainControl.java#L61

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+                 sample_rate=16000,
+                 frame_width=20,
+                 target_level_dbfs=0,
```
brentspell, line 30: `target_level_dbfs` and `compression_gain_db` should have defaults that are consistent with android
https://github.com/spokestack/spokestack-android/blob/master/src/main/java/io/spokestack/spokestack/webrtc/AutomaticGainControl.java#L45

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+        self._frame_width = frame_width
+        self._target_level_dbfs = target_level_dbfs
+        self._compression_gain_db = compression_gain_db
+        self._limit_enable = limit_enable
+
+        self._agc = NULL
+        result = cagc.WebRtcAgc_Create(&self._agc)
+        if result < 0:
+            raise ValueError
+        if result == 0:
```
brentspell, line 45: this if is not needed, since you are throwing on an invalid return code in the previous statement

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+        self._sample_rate = sample_rate
+        self._frame_width = frame_width
+        self._target_level_dbfs = target_level_dbfs
+        self._compression_gain_db = compression_gain_db
+        self._limit_enable = limit_enable
+
+        self._agc = NULL
+        result = cagc.WebRtcAgc_Create(&self._agc)
+        if result < 0:
+            raise ValueError
```
brentspell, line 44: this should be `MemoryError`, since the only way it can fail is if an allocation fails
also, the check above should be `result != 0` because 0 is the only successful return code

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+                                         maxLevel=MIC_MAX,
+                                         agcMode=2,
+                                         fs=self._sample_rate)
+
+            if result == 0:
+                self._config.targetLevelDbfs = self._target_level_dbfs
+                self._config.limiterEnable = self._limit_enable
+                self._config.compressionGaindB = self._compression_gain_db
+                result = cagc.WebRtcAgc_set_config(self._agc, self._config)
+        if result < 0:
```
brentspell, line 57: same here, want to check `result != 0`, and I would add a symbolic message to that `ValueError` (`invalid_config` or whatever)

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+                inNearH=NULL,
+                samples=<short> len(frame),
+                out=<short*> np.PyArray_DATA(frame),
+                out_H=NULL,
+                inMicLevel=MIC_TARGET,
+                outMicLevel=&mic_level,
+                echo=0,
+                saturationWarning=&saturated
+            )
+        if result < 0:
```
brentspell, line 87: same here, also ValueError probably doesn't make sense here, since this is not a parameter error
you may want to define a custom exception for this (maybe `ProcessError`)

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,94 @@
...
+            if result == 0:
+                self._config.targetLevelDbfs = self._target_level_dbfs
+                self._config.limiterEnable = self._limit_enable
+                self._config.compressionGaindB = self._compression_gain_db
+                result = cagc.WebRtcAgc_set_config(self._agc, self._config)
+        if result < 0:
+            raise ValueError
+
+    def __call__(self, context, frame):
+        self._process(frame)
```
brentspell, line 61: the frame length and dtype need to be validated here
otherwise, you can crash the client application if they give you the wrong sized frame, wrong sample rate, etc.
there should also be tests to verify an invalid frame length/dtype

spokestack/extensions/webrtc/ans.pyx
```diff
@@ -0,0 +1,59 @@
...
+# distutils: include_dirs = filter_audio/ns/include/
+cimport cans
+cimport numpy as np
+
+
+np.import_array()
+
+cdef class AutomaticNoiseSuppression:
+    cdef cans.NsxHandle *_ans
+    cdef int _sample_rate
```
brentspell, line 14: `_sample_rate` and `_policy` are not used outside the constructor, so they don't need to be stored

spokestack/extensions/webrtc/ans.pyx
```diff
@@ -0,0 +1,59 @@
...
+    cdef int _policy
+    cdef int _frame_width
+
+    def __dealloc__(self):
+        cans.WebRtcNsx_Free(self._ans)
+        self._ans = NULL
+
+    def __init__(self, sample_rate=16000, policy=0, **kwargs):
```
brentspell, line 22: the configuration parameter validation needs to be ported from android
https://github.com/spokestack/spokestack-android/blob/master/src/main/java/io/spokestack/spokestack/webrtc/AcousticNoiseSuppressor.java#L61

brentspell, line 22: the policy parameter should be something more friendly to the client
consider creating symbolic constants for these like the vad does

spokestack/extensions/webrtc/ans.pyx
```diff
@@ -0,0 +1,59 @@
...
+        self._ans = NULL
+        self._sample_rate = sample_rate
+        self._policy = policy
+        self._frame_width = self._sample_rate * 10 / 1000
+
+        if self._frame_width % 10 != 0:
+            raise ValueError
+
+        result = cans.WebRtcNsx_Create(&self._ans)
+        if result == 0:
```
brentspell, line 32: we are silently ignoring failure here and in the check below
failures in these checks should result in exceptions

spokestack/extensions/webrtc/ans.pyx
```diff
@@ -0,0 +1,59 @@
...
+            self._process(frame[i:frame_size + i])
+
+    cdef _process(self, frame):
+        result = cans.WebRtcNsx_Process(self._ans,
+                                        <short*> np.PyArray_DATA(frame),
+                                        NULL,
+                                        <short*> np.PyArray_DATA(frame),
+                                        NULL)
+        if result != 0:
+            raise ValueError
```
brentspell, line 53: I would create a custom exception for this

spokestack/extensions/webrtc/vad.pyx
```diff
@@ -0,0 +1,41 @@
...
+
+    def __dealloc__(self):
+        cvad.WebRtcVad_Free(self._vad)
+
+    def __init__(self, sample_rate=16000, mode=0):
+        self._sample_rate = sample_rate
+        self._mode = mode
+
+        result = cvad.WebRtcVad_Create(&self._vad)
+        if result == 0:
```
brentspell, line 20: we are silently ignoring failure in all of these `if result` checks
failures in these checks should result in exceptions

spokestack/extensions/webrtc/vad.pyx
```diff
@@ -0,0 +1,41 @@
...
+
+cdef class WebRtcVad:
+    cdef cvad.VadInst*_vad
+    cdef int _sample_rate
+    cdef int _mode
+
+    def __dealloc__(self):
+        cvad.WebRtcVad_Free(self._vad)
+
+    def __init__(self, sample_rate=16000, mode=0):
```
brentspell, line 15: this constructor needs to validate its configuration parameters like the android version
https://github.com/spokestack/spokestack-android/blob/master/src/main/java/io/spokestack/spokestack/webrtc/VoiceActivityDetector.java#L80

spokestack/extensions/webrtc/ans.pyx
```diff
@@ -0,0 +1,59 @@
...
+
+        result = cans.WebRtcNsx_Create(&self._ans)
+        if result == 0:
+            result = cans.WebRtcNsx_Init(self._ans, self._sample_rate)
+
+            if result == 0:
+                result = cans.WebRtcNsx_set_policy(self._ans, self._policy)
+
+    def __call__(self, context, frame):
+        frame_size = self._frame_width
```
brentspell, line 39: I would also check the dtype here

spokestack/extensions/webrtc/ans.pyx
```diff
@@ -0,0 +1,59 @@
...
+        if len(frame) % frame_size != 0:
+            raise ValueError
```
brentspell, line 41: this exception could use a symbolic message ("invalid_frame_length" or whatever)

spokestack/extensions/webrtc/vad.pyx
```diff
@@ -0,0 +1,41 @@
+cimport cvad
+cimport numpy as np
+
+
+np.import_array()
+
+cdef class WebRtcVad:
+    cdef cvad.VadInst*_vad
+    cdef int _sample_rate
```
brentspell, line 9: the `_sample_rate` and `_mode` members are only used in the constructor, so they don't need to be stored

spokestack/vad/webrtc.py
```diff
@@ -1,13 +1,12 @@
 """
-This module contains the webrtc component for
-voice activity detection (vad)
+This module contains the webrtc component for voice activity detection (vad)
```
brentspell, line 2: it is a bit weird to have a separate wrapper module/class for the vad stuff and not for the other webrtc components
since the client won't see it, I suppose this is fine, but if you can find a way to make them all consistent, that would be ideal

will-rice: yeah. that makes sense. I definitely like the way the vad is setup vs the other extensions.

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,81 @@
...
+                 sample_rate,
+                 frame_width,
+                 target_level_dbfs,
+                 compression_gain_db,
+                 limit_enable
+    ):
+        self._agc = NULL
+        result = cagc.WebRtcAgc_Create(&self._agc)
+        if result != 0:
+            raise ValueError("invalid_config")
```
brentspell, line 31: there is no configuration being passed here, so this should be MemoryError, since that's the only way it can fail

spokestack/extensions/webrtc/agc.pyx
```diff
@@ -0,0 +1,81 @@
...
+
+        cdef cagc.WebRtcAgc_config_t config
+        config.targetLevelDbfs = target_level_dbfs
+        config.limiterEnable = limit_enable
+        config.compressionGaindB = compression_gain_db
+
+        result = cagc.WebRtcAgc_set_config(self._agc, config)
+        if result != 0:
+            raise MemoryError
```
brentspell, line 49: this should probably be an invalid config error, since you're passing it configuration parameters, and allocations were probably done in create

spokestack/extensions/webrtc/nsx.pyx
```diff
@@ -0,0 +1,44 @@
...
+    def __dealloc__(self):
+        cnsx.WebRtcNsx_Free(self._ans)
+        self._ans = NULL
+
+    def __init__(self, sample_rate, policy):
+        self._ans = NULL
+
+        result = cnsx.WebRtcNsx_Create(&self._ans)
+        if result != 0:
+            raise ValueError("invalid_config")
```
brentspell, line 24: MemoryError here

spokestack/extensions/webrtc/nsx.pyx
```diff
@@ -0,0 +1,44 @@
...
+
+        result = cnsx.WebRtcNsx_Init(self._ans, sample_rate)
+        if result != 0:
+            raise ValueError("invalid_config")
+
+        result = cnsx.WebRtcNsx_set_policy(self._ans, policy)
+        if result != 0:
+            raise MemoryError
```
brentspell, line 32: invalid config here

spokestack/extensions/webrtc/vad.pyx
```diff
@@ -0,0 +1,37 @@
...
+    def __dealloc__(self):
+        cvad.WebRtcVad_Free(self._vad)
+
+    def __init__(self, sample_rate, mode):
+        self._vad = NULL
+        self._sample_rate = sample_rate
+
+        result = cvad.WebRtcVad_Create(&self._vad)
+        if result != 0:
+            raise ValueError("invalid_config")
```
brentspell, line 20: memory error

spokestack/extensions/webrtc/vad.pyx
```diff
@@ -0,0 +1,37 @@
...
+
+        result = cvad.WebRtcVad_Init(self._vad)
+        if result != 0:
+            raise ValueError("invalid_config")
+
+        result = cvad.WebRtcVad_set_mode(self._vad, mode)
+        if result != 0:
+            raise MemoryError
```
brentspell, line 28: invalid config

spokestack/nsx/webrtc.py
```diff
@@ -0,0 +1,57 @@
...
+POLICY_AGGRESSIVE = 2
+POLICY_VERY_AGGRESSIVE = 3
+
+
+class AutomaticNoiseSuppression:
+    """WebRTC Automatic Noise Suppression
+
+    Args:
+        sample_rate (int): audio sample rate. (Hz)
+        policy (int): level of noise supression.
```
brentspell, line 20: policy could probably use some better documentation here
see what the sphinx docs look like and consider how useful they would be to a new user

brentspell (commented):
pretty close, just a couple of exception inconsistencies and docstrings

PR #45, opened 2021-01-29
Shield experiment on README.md

roppem9:
@noelweichbrodt didn't realize that you can customize shields on repos (thanks, Will!) 

What do you think? I was able to use our colors too. Looks cleaner to us...


noelweichbrodt (approved):
Nice!

PR #46, opened 2021-01-30
Readthedocs Build Update

will-rice:
This commit fixes various build issues with the autodocs. One change was also made to setup.py based on a readthedocs [recommendation](https://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules)


brentspell (commented):
lgtm

PR #48, opened 2021-02-01
Add Dependencies to Publish Script

will-rice:
This adds numpy and Cython to the publish workflow. The on-release failed because numpy wasn't installed.


brentspell:
lgtm

PR #49, opened 2021-02-01
Added Submodule Init to Workflow

will-rice:
I believe this is the only discrepancy remaining.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -4,27 +4,27 @@
 name: Upload Python Package
 
 on:
-    release:
-        types: [created]
+  release:
+    types: [created]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -4,27 +4,28 @@
 name: Upload Python Package
 
 on:
-    release:
```
brentspell, line 7: See if you can figure out a way to make this change without reformatting the whole file. If you have to reformat it, do that on a separate commit.

will-rice: I believe I fixed it. it was pre-commit.

brentspell:
nice, lgtm

PR #51, opened 2021-02-08
Add start call to run if pipeline is not running

will-rice:


tests/test_pipeline.py
```diff
@@ -70,11 +70,6 @@ def test_cleanup():
     pipeline.stop()
     assert not pipeline.is_running
 
-    # run after stopped will trigger clean up
```
space-pope, line 73: This method doesn't test cleanup anymore—and `test_start_stop` above doesn't call stop. It would probably be best to roll the two into one test.

PR #54, opened 2021-02-26
Remove setting the input source to None in cleanup.

will-rice:
We still want to call clean up in `run`, but if we set the input source to `None` the pipeline can't be started again.


brentspell:
lgtm

PR #58, opened 2021-03-08
Allow Pipeline to Restart After Stopped

will-rice:
Initially, the pipeline set the input source to None on a call to stop. This blocked the ability to start again without creating a new pipeline instance. There are certain situations where the user may want to halt the pipeline for a period longer than a pause, but stil be able to start again without restarting the main application.


brentspell:
lgtm

PR #59, opened 2021-03-12
Refactor setup.py and publish workflow

will-rice:
This change includes building [PortAudio](http://www.portaudio.com/) for the specific platform on install. Prior to this, the user had to install PortAudio from the system package manager before installing PyAudio. Now, we can ship with Pyaudio included and give users microphone support just from the pip install. Another part of this refactor, introduced the ability to build multi-platform wheels through [cibuildwheel](https://github.com/joerick/cibuildwheel) This should make installing spokestack faster/easier for all platforms.


brentspell:
This will require maintaining and distributing updates to portaudio as it changes to support new/changing platforms, which happens regularly, but I suppose that avoiding platform-specific installation steps is worth it.

PR #61, opened 2021-03-15
Pin numpy version to 1.19.2 resolves ValueError issue

will-rice:
Pinning the python version to `1.19.2` fixes the `ValueError`. This is also the version required for the last couple TensorFlow releases.


brentspell:
I'm unsubbing from this one for now, but ping me when it's ready to go.

PR #65, opened 2021-03-17
Add wakeword SpeechPipeline profile.

will-rice:
This change adds a plain wakeword profile for cases where the user may want to use only the wakeword without asr.

### DIFF
--- a/spokestack/profile/wakeword.py
+++ b/spokestack/profile/wakeword.py
@@ -17,10 +17,7 @@ class SpokestackWakeword:
 
     @staticmethod
     def create(
-        sample_rate: int = 16000,
-        frame_width: int = 20,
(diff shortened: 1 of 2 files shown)


spokestack/profile/wakeword.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class SpokestackWakeword:
+    """ Spokestack wakeword profile. """
+
+    @staticmethod
+    def create(
+        sample_rate: int = 16000,
+        frame_width: int = 20,
+        model_dir: str = "",
```
brentspell, line 22: this profile will fail immediately if `model_dir` isn't specified, so it shouldn't be an optional parameter

brentspell:
lgtm

PR #66, opened 2021-03-26
Refactor GoogleSpeechRecognizer to support >=3.6

will-rice:
This change enables backward compatibility for Python 3.7.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ setup(
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: OS Independent",
     ],
-    python_requires=">=3.8",
+    python_requires=">=3.6",
     setup_requires=["setuptools", "wheel", "numpy==1.19.2", "Cython>=0.29.22"],
     install_requires=[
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -123,7 +123,7 @@ def run(self) -> None:
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: OS Independent",
     ],
-    python_requires=">=3.8",
+    python_requires=">=3.7",
```
space-pope, line 126: Does it still require new features introduced in 3.7?

space-pope (commented):
LGTM if it works without that initial blocking `get`. Also had a question about compatibility.

setup.py
will-rice: I don’t think so, but that’s something I will test. I'll see how far back we can go. There was also a build issue which doesn’t make a lot of sense so investigating that as well.

space-pope (approved):
LGTM. Thanks for widening the net!

PR #67, opened 2021-03-27
Update pre-commit-config.yml with the latest versions of hooks.

will-rice:
This change updated pre-commit hooks and then adds the formatting changes to .git-blame-ignore-revs.


brentspell:
lgtm

PR #68, opened 2021-03-29
Update README.md

will-rice:
The first commit updates the formatting and the second commit adds the "Build Extensions" directions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ pipeline = WakewordSpokestackASR.create(
 
 ### Speech Pipeline
 
-If you would like fine-grained control over what is included in the pipeline, you can use `SpeechPipeline`.This is the module that ties together VAD (voice activity detection), wakeword, and ASR (automated speech detection). The VAD listens to a frame of audio captured by the input device to determine if speech is present. If it is, the wakeword model processes subsequent frames of audio looking for the keyword it has been trained to recognize. If the keyword is found, the pipeline is activated and performs speech recognition, converting the subsequent audio into a transcript. The `SpeechPipeline` is initialized like this:
+If you would like fine-grained control over what is included in the pipeline, you can use `SpeechPipeline`. This is the module that ties together VAD (voice activity detection), wakeword, and ASR (automated speech detection). The VAD listens to a frame of audio captured by the input device to determine if speech is present. If it is, the wakeword model processes subsequent frames of audio looking for the keyword it has been trained to recognize. If the keyword is found, the pipeline is activated and performs speech recognition, converting the subsequent audio into a transcript. The `SpeechPipeline` is initialized like this:
 
 ```python
 from spokestack.activation_timeout import ActivationTimeout
@@ -92,7 +92,6 @@ timeout = ActivationTimeout()
 
 
 pipeline = SpeechPipeline(mic, [vad, wake, asr, timeout])
-pipeline.start()
 pipeline.run()
 ```
 
--- a/README.md
+++ b/README.md
@@ -54,6 +54,14 @@ pyenv local spokestack
 pip install -r requirements.txt
 ```
 
+### Build Extensions
+
+When installing from source, you will want to build the Cython extensions.
+
+```shell
+python setup.py build_ext --inplace
+```
+
 ### Install Tensorflow
 
 This library requires a way to run [TFLite](https://www.tensorflow.org/lite) models. There are two ways to add this ability. The first is installing the full [Tensorflow](https://www.tensorflow.org/) library.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,6 +54,14 @@ pyenv local spokestack
 pip install -r requirements.txt
 ```
 
+### Build Extensions
+
+When installing from source, you will want to build the Cython extensions.
```
space-pope, line 59: It's not necessarily clear that you have to install to run the tests. This might be a better fit for a setup section in the CONTRIBUTING guide or a more general "development" section in the readme.

As it is, this section mixes setup for use (`pip install`) with setup for development on the library itself, which are two pretty different use cases.

You'll also want to add a note about the submodules; you have to run `git submodule update --init` or the equivalent before the extensions can be built.

space-pope (changes requested):
The potential restructuring is up to you, but I do think the submodule note needs to be there.

README.md
will-rice: This sounds good. I will move the development specific items to CONTRIBUTING

will-rice: I did some rearranging also and put `pip install spokestack` at the very top.

README.md
```diff
@@ -9,9 +9,31 @@ Welcome to Spokestack Python! This library is intended for developing [voice int
...
+
+The full Tensorflow package is installed with the following:
+
+```shell
+pip install tensorflow
+```
+
+#### TFLite Interpreter (Embedded Devices)
+
+In use cases where you require a small footprint, such as on a Raspberry Pi or similar embedded devices, you will want to install the TFLite Interpreter. You can install it for your platform by [following the instructions](https://www.tensorflow.org/lite/guide/python#install_just_the_tensorflow_lite_interpreter).
```
space-pope, line 32: Why not include the `pip install` command they list, just for the convenience of not clicking away (and to match the previous section)?

README.md
```diff
@@ -35,46 +57,27 @@ Another potential avenue for using `spokestack` on Windows 10 is from [anaconda]
...
-
 ### Speech Pipeline
 
-The Speech Pipeline is the module that ties together VAD (voice activity detection), wakeword, and ASR (automated speech detection). The VAD listens to a frame of audio captured by the input device to determine if speech is present. If it is, the wakeword model processes subsequent frames of audio looking for the keyword it has been trained to recognize. If the keyword is found, the pipeline is activated and performs speech recognition, converting the subsequent audio into a transcript. The Speech Pipeline is initialized like this:
+If you would like fine-grained control over what is included in the pipeline, you can use `SpeechPipeline`.This is the module that ties together VAD (voice activity detection), wakeword, and ASR (automated speech detection). The VAD listens to a frame of audio captured by the input device to determine if speech is present. If it is, the wakeword model processes subsequent frames of audio looking for the keyword it has been trained to recognize. If the keyword is found, the pipeline is activated and performs speech recognition, converting the subsequent audio into a transcript. The `SpeechPipeline` is initialized like this:
```
space-pope, line 77:
```suggestion
If you would like fine-grained control over what is included in the pipeline, you can use `SpeechPipeline`. This is the module that ties together VAD (voice activity detection), wakeword, and ASR (automated speech detection). The VAD listens to a frame of audio captured by the input device to determine if speech is present. If it is, the wakeword model processes subsequent frames of audio looking for the keyword it has been trained to recognize. If the keyword is found, the pipeline is activated and performs speech recognition, converting the subsequent audio into a transcript. The `SpeechPipeline` is initialized like this:
```

README.md
```diff
@@ -85,9 +88,10 @@ mic = PyAudioInput()
 vad = VoiceActivityDetector()
 wake = WakewordTrigger("path_to_tflite_model")
 asr = SpeechRecognizer("spokestack_id", "spokestack_secret")
+timeout = ActivationTimeout()
 
 
-pipeline = SpeechPipeline(mic, [vad, wake, asr])
+pipeline = SpeechPipeline(mic, [vad, wake, asr, timeout])
```
space-pope, line 94: In newer versions, the `start()` call below isn't needed, right?

space-pope (commented):
Had a couple extra comments here that I had written for your original PR, but then removed when the formatting change got reverted.

README.md
will-rice: yeah I can remove that also

README.md
will-rice: yep. thanks for pointing that out again. might as well get it with this one also.

space-pope (approved):
LGTM. Thanks for the extra updates!

PR #69, opened 2021-03-29
Expose TTS profile to clients

space-pope:
This change allows clients to access the `profile` argument
for the GraphQL synthesis method so that audio formatted for
third-party services can be produced. It also condenses the
GraphQL query creation and adds some documentation for the
new option.

Default values have been added to the manager's `synthesize`
method to support some example code from the readme. The
enclosing GraphQL query has also been changed to include the
platform name.

### DIFF
--- a/spokestack/tts/clients/spokestack.py
+++ b/spokestack/tts/clients/spokestack.py
@@ -10,7 +10,7 @@ from typing import Any, Iterator
 import requests
 
 _MODES = {
-    "ssml": "synthesizeSSML",
+    "ssml": "synthesizeSsml",
     "markdown": "synthesizeMarkdown",
     "text": "synthesizeText",
 }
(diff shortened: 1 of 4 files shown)


spokestack/tts/clients/spokestack.py
```diff
@@ -10,7 +10,7 @@
 import requests
 
 _MODES = {
-    "ssml": "synthesizeSSML",
+    "ssml": "synthesizeSsml",
```
space-pope, line 13: Interestingly enough, this appears to work both ways, but this version is more consistent with the other platforms and the formatting in the GraphQL schema.

spokestack/tts/clients/spokestack.py
```diff
@@ -94,48 +124,27 @@ def synthesize_url(
...
-                    "query": """
-                           query synthesis($voice: String!, $text: String!) {
-                             synthesizeText(voice: $voice, text: $text) {url}
-                           }
-                           """,
-                    "variables": {"voice": voice, "text": message},
-                }
-            )
-        else:
+    def _build_body(message: str, mode: str, voice: str, profile: str) -> str:
```
will-rice, line 127: Really nice refactor here.

will-rice (approved):
lgtm

PR #70, opened 2021-03-31
Fix Copyright Year for README and Readthedocs

will-rice:
I auto-pulled the year for readthedocs, but I'm not sure how to do it for the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ twine upload dist/*
 
 ## License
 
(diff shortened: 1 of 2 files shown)


will-rice:
I saw some javascript ways to do it, but I don't know if we want to add javascript to the readme.

space-pope:
Nah. It only happens once a year :).

PR #71, opened 2021-03-31
Mock extensions in sphinx-build

will-rice:
The modules that require Cython extensions were not being built for readthedocs.


space-pope (approved):
Awesome; glad this was an easy fix.

PR #75, opened 2021-04-15
Add sounddevice as an audio input source.

will-rice:
This change introduces `SoundDeviceInput` to the `io` module. Sounddevice is now the preferred method of capturing audio input, due to better multi-platform distribution support than PyAudio. However, PyAudio will continue to be supported for the forseeable future.


spokestack/io/sound_device.py
```diff
@@ -0,0 +1,73 @@
...
+class SoundDeviceInput:
+    """
+    Audio input source using sounddevice.
+
+    Parameters
+    ----------
+    sample_rate : int, optional
+        audio sample rate, by default 16000
+    frame_width : int, optional
+        size of the audio frame, by default 20
```
space-pope, line 17:
```suggestion
        size of the audio frame in milliseconds, by default 20
```
I'm used to seeing "default: <value>" or "defaults to <value>" rather than "by default", but that's a hardcore nitpick.

space-pope (approved):
LGTM, nice addition.

Weird that the commit hook didn't catch those quotes in the test the first time around. Technically that should be in a separate commit/PR, but no need to go back and change it.

will-rice:
I think `pre-commit` has trouble with nested directories. it would habitually miss the docs.

PR #87, opened 2021-05-04
add tts-lite components

brentspell:
These changes add a new TensorFlow-Lite speech synthesizer for generating and playing speech locally.

A sample model that can be used for example projects/blog posts/etc. can be found at https://s.spokestack.io/u/aKvWb/model.tar.gz.


will-rice (approved):
lgtm. this is a really big deal. nice work.

PR #90, opened 2021-05-06
Update issue templates

will-rice:
Adding these default issue templates. They look good enough for a start, but we may want to customize with specifics later.


brentspell:
I don't think I have much feedback to offer here, but this should probably be made consistent with whatever is being done on the other platforms
josh may have some better feedback for you on this

will-rice:
ok sorry, I will talk to him.

PR #100, opened 2021-06-30
Add PyTorch model abstraction

will-rice:
PyTorch is a popular ML framework in Python. Users may want to be able to use jit traced PyTorch models with Spokestack. This abstraction offers users that ability.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,7 @@ jobs:
           pip install --upgrade pip pip-tools
           pip install -r requirements.txt
           pip install tensorflow
+          pip install torch
           python -m spacy download en_core_web_sm
       - save_cache:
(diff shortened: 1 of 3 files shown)


space-pope (commented):
This allows PyTorch models to be loaded, but doesn't _do_ much with them. The readme instructions explain how to load a model, but not how to use one in the context of Spokestack. Are further integrations/instructions planned for future PRs?

In other words, this is more of a library contributor/developer feature as written, something that you'd mention in tutorials about building your own integrations rather than the readme.

will-rice:
This is intended to be the complement to the TFLite model abstraction. Ultimately, I think we could let the user choose between TensorFlow or PyTorch. I will take out the documentation in the README, because like you said only a contributor really needs to know about this class.

space-pope (approved):

